PR #2, opened 2014-12-24
Always send a referer but spoof it first

fmarier:
Blocking referers breaks some sites, but these sites will happily accept a
spoofed one.

### DIFF
--- a/user.js
+++ b/user.js
@@ -38,8 +38,10 @@
 // https://www.mozilla.org/projects/security/pki/nss/nss-3.11/nss-3.11-algorithms.html
 
 // http://kb.mozillazine.org/Network.http.sendRefererHeader#0
-// Never send the Referer header or set document.referrer.
-user_pref("network.http.sendRefererHeader",	0);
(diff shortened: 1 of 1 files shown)


pyllyukko:
Yeah, I think we can make this the default. As this still enables few more sites and makes browsing a bit less of a hassle.

BTW. I noticed that [IMDb](http://www.imdb.com/)'s login screen does the checking also and requires also ```network.http.sendRefererHeader == 2```.

Thanks.

CHEF-KOCH:
The following plugins are able to take control over the referers, _refControl_ and _Random Agent Spoofer_. Both using normally "default  (on)" on every page. Personally I think we should remove that option in the user.js script and control such behavior with the given plugins, which allow us to control that per-site only and not only global.

pyllyukko:
I think @fmarier had good reasoning for this and this is something that doesn't necessarily need any changing. Now it sends it and spoofs it, which works quite nicely and doesn't leak information. I've been thinking of changing the ```network.http.sendRefererHeader == 2``` to fix the few logins that require it.

PR #6, opened 2014-12-25
Replaced a few TODOs by the variables they were refering to, updated the README

lenormf:
````
I noticed a few links documenting variables that were not being set by the userscript, so I just added them in. I have not extensively tested my modification, although it doesn't seem to have broken anything so far.
````
Second pull request with the same content splitted into several commits.


pyllyukko:
I'll close this pull request now, as the few missing variables still need verification (also, it can't be automatically merged anymore). I created an issue (#9) for these settings.

Thanks for the contribution!

PR #15, opened 2015-05-02
Update README.md

CHEF-KOCH:
* added Shumway (a secure Adobe Flash alternative HTML5 based - works quite good on YT/Vimeo/...)
* added Random Agent Spoofer which offers a GUI about the most important leak stuff like timezone, webgl, and such ...
* added Tor's default search engine, it doesn't require JavaScript or Captchas - sure DuckDUck also have a no JavaScript version but after x searches it requires to enter a Captcha - it also combines DuckDuck, Bing, Google,.. search results
* added RCC certificate tool, it's quite helpful because it scans for default certificates and show if there are certificates that doesn't belong to Windows/Firefox/CyberFox/WaterFox/,...
* added link to Mozilla's default CA list (I not linked in the .txt directly since it may could be changed in the future).

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,13 +147,16 @@ Here is a list of the most essential security and privacy enhancing add-ons that
   * See also: [Are We Private Yet?](http://www.areweprivateyet.com/) site
 * [Privacy Badger](https://www.eff.org/privacybadger)
 * [DuckDuckGo Plus](https://addons.mozilla.org/en-US/firefox/addon/duckduckgo-for-firefox/) (instead of Google)
+  * Similar but without Captchas [Disconnect Search](https://search.disconnect.me/)
 * [Mozilla Lightbeam](https://www.mozilla.org/en-US/lightbeam/)
+* [Mozilla Shumway](https://github.com/mozilla/shumway/) (Adbobe Flash replacement)
+* [Random Agent Spoofer](https://github.com/dillbyrne/random-agent-spoofer)
 
 Online tests
 ------------
 
 * [Panopticlick](https://panopticlick.eff.org/)
-* [www.filldisk.com](http://www.filldisk.com/)
+* [FillDisk](http://www.filldisk.com/)
 * [SSL Client Test](https://www.ssllabs.com/ssltest/viewMyClient.html)
 * [evercookie](http://samy.pl/evercookie/)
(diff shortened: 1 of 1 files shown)


pyllyukko:
Are there sources available for that RCC tool? Also, we're not talking about Mozilla's default CA list, but our script's custom CA list.

pyllyukko:
Could you split this into smaller chunks? I'd like to merge parts of it, but not all at least not yet.

CHEF-KOCH:
> Also, we're not talking about Mozilla's default CA list, but our script's custom CA list.

Sure but Mozilla's list is almost the same (I guess except 2 certificates) and should be preferred since it's trusted/based on compared to others. See [here](http://mxr.mozilla.org/mozilla-central/source/security/nss/lib/ckfw/builtins/certdata.txt?raw=1) for the source. Microsoft itself also offers such checks via [sigcheck -q -e -s -c ](https://technet.microsoft.com/en-us/sysinternals/bb897441.aspx) which works with Firefox too. 

Sad RCC isn't open source (_yet_), but [I asked](http://www.wilderssecurity.com/threads/rcc-check-your-systems-trusted-root-certificate-store.373819/page-2#post-2481366) for it on the _official_? [wilderssecurity forum thread](http://www.wilderssecurity.com/threads/rcc-check-your-systems-trusted-root-certificate-store.373819/). If someone not trust or need the source I would say that we can use Microsoft's script and change to our needs.

Maybe I split it, but I already work on the next version which also explain and fix some more stuff, especially the fingerprinting stuff.

pyllyukko:
> Also, we're not talking about Mozilla's default CA list, but our script's custom CA list.
> 
> Sure but Mozilla's list is almost the same (I guess except 2 certificates) and should be preferred since it's trusted/based on compared to others. See here for the source. Microsoft itself also offers such checks via sigcheck -q -e -s -c which works with Firefox too.

Huh?

```
$ grep "^# Certificate" certdata.txt |wc -l
203
```

Firefox's default CA list packs a whole bunch and our list is down to 24. That was the whole point to reduce from the Firefox's built-in list. I don't follow.

pyllyukko:
I'll close this now, as the master has already progressed beyond automatic merging.

PR #18, opened 2015-05-21
logjam related disabling of two more DH ciphers

graste:
See "The logjam attack" – https://weakdh.org/. The two ciphers are enabled at the moment in user.js but are susceptible to the logjam attack and thus should be disabled as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,8 +119,6 @@ Here's the list with this config:
 	    Cipher Suite: TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 (0xc02f)
 	    Cipher Suite: TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA (0xc00a)
 	    Cipher Suite: TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA (0xc014)
-	    Cipher Suite: TLS_DHE_RSA_WITH_AES_256_CBC_SHA (0x0039)
-	    Cipher Suite: TLS_DHE_RSA_WITH_CAMELLIA_256_CBC_SHA (0x0088)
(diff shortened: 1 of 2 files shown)


graste:
Hm, now that I scroll through the README these two entries are no longer valid with the acceptance of this PR, aren't they?
```
    Cipher Suite: TLS_DHE_RSA_WITH_AES_256_CBC_SHA (0x0039)
    Cipher Suite: TLS_DHE_RSA_WITH_CAMELLIA_256_CBC_SHA (0x0088)
```
Should I make a second commit and remove them from the README file?

pyllyukko:
Indeed we could mitigate against LogJam with this.

I'm afraid many sites will break if we disable these two. Have you tried whether you're able to browse freely with these disabled?

I'll switch them off myself and check back whether it causes too much issues.

pyllyukko:
> Should I make a second commit and remove them from the README file?

And yes, the README should be updated also. Second commit should do the trick.

graste:
You're right. I didn't try it. Will change the README though in case no problems appear and the PR might be acceptable…

pyllyukko:
Please test with your browser also and tell us if there's any issues? Thanks.

graste:
Will do, but it's not my everyday browser at the moment. :-)

pyllyukko:
With (very) quick testing, no sites found... will continue.

pyllyukko:
Ok. I'll merge this. If there's some problems, we'll just revert it. It's a shame, as these ciphers have the [forward secrecy](https://en.wikipedia.org/wiki/Forward_secrecy) property. I think we should re-enable them at some point when the dust settles.

PR #23, opened 2015-06-28
Update user.js

CHEF-KOCH:
* New file structure -> Description, See, Note and additional Notes and Comments 
* Fixed some settings in the name of security 
* Added a own Deprecated section to store obsolete settings in there, in the future this will be sort FF35,36,37,38,.. so that everyone can see what was changed an improved, it's like a history but if someone use an older version he still can uncomment the specific settings
* Character encoding of this document need to be checked (not fully done) e.g. the spaces between ",			true); and so are not always consistent 
* Some related settings added e.g. browser.sessionstore.privacy_level_deferred but they are comment out because that doesn't affect everyone because it's depending on which settings are used 
* Typos and spelling fixed

### DIFF
--- a/user.js
+++ b/user.js
@@ -1,333 +1,336 @@
 /******************************************************************************
  * user.js                                                                    *
- * https://github.com/pyllyukko/user.js                                       *
+ * https://github.com/pyllyukko/user.js										  *
+ * 																			  *
+ * This is the Firefox own user.js which stores the changed about:config      *
+ * settings     															  *
  ******************************************************************************/
 
+
+
 /******************************************************************************
- * HTML5 / APIs / DOM                                                         *
+ * HTML5 / APIs / DOM related                                                 *
  *                                                                            *
  ******************************************************************************/
 
-// disable Location-Aware Browsing
-// http://www.mozilla.org/en-US/firefox/geolocation/
-user_pref("geo.enabled",			false);
+// Description: Disable Location-Aware Browsing
+// See: 		http://www.mozilla.org/en-US/firefox/geolocation/
+user_pref("geo.enabled",		false);
 
-// http://kb.mozillazine.org/Dom.storage.enabled
-// http://dev.w3.org/html5/webstorage/#dom-localstorage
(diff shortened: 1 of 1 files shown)


pyllyukko:
That's a lot of changes! I really can't merge this in it's current form, as it would be important to see for instance changed settings in the Git log, etc.

For instance:
> Fixed some settings in the name of security

These would have to be in separate commits with the reasoning preferably in the commit message.

> Added a own Deprecated section to store obsolete settings in there, in the future this will be sort FF35,36,37,38,.. so that everyone can see what was changed an improved, it's like a history but if someone use an older version he still can uncomment the specific settings

Even though some setting are the same by default, it is intended to have them explicitly set in the ```user.js``` file, as explained in the "Some of the settings in this user.js file might seem redundant" part of the README.

> Character encoding of this document need to be checked (not fully done) e.g. the spaces between ", true); and so are not always consistent

I have used tabs to make e.g. ```grep ^user_pref user.js``` view consistent, even though the amount of tabs currently varies a bit.

I see that you have done a great deal of work with this, and I appreciate your contribution, but I can't have so many changes in one commit. Is it possible to have this in as small pieces as possible? Preferably even separate pull requests for different aspects?

CHEF-KOCH:
Yes, I can separate these things you mentioned. But I'm not happy with these since this coast more time. - My time. 
I'm almost done with the Firefox support, after that I can submit something new. In fact we need to discuss if we merge the settings with the mobile version, but as you know I'm not a fan of depcreated settings since my philosophy is to always (or most time) stay up2-date especially for something that I use and need every day.

Until after all is done I close this now, if anyone want's to help feel free to use this as a base.

pyllyukko:
> Yes, I can separate these things you mentioned. But I'm not happy with these since this coast more time. - My time.

I realize that, and that's why I feel bad to say no.

> In fact we need to discuss if we merge the settings with the mobile version

You can also contact me at IRC (OFTC & Freenode) to discuss more about these.

Issue #32, opened 2015-08-09
Duplicated settings


Genbox:
The following settings are duplicated in the config file:
security.ssl3.dhe_dss_des_ede3_sha
security.ssl3.ecdh_ecdsa_des_ede3_sha
security.ssl3.ecdh_rsa_des_ede3_sha
security.ssl3.rsa_aes_128_sha
security.ssl3.rsa_aes_256_sha

The first 3 seem to be duplicated because they fulfill multiple criteria to be disabled. The fourth cipher on the list is commented out, but then enabled again later on. The last cipher is first disabled, then enabled again.

Consider removing duplicates to simplify the configuration file and remove conflicting settings.

CHEF-KOCH:
Thanks for reporting this, but instead of opening several threads it would be easier to make a pull request. I will grep all your issues now and make a pull.

ian-csis:
Sorry about that. I would have made a pull request if I knew beforehand that I would find several changes/bugs. However, I started work on the configuration website as mentioned in #25 and found the issues as development progressed.

Genbox:
Sorry about that. I would have made a pull request if I knew beforehand that I would find several changes/bugs. However, I started work on the configuration website as mentioned in #25 and found the issues as development progressed.

pyllyukko:
Merged

Issue #30, opened 2015-08-08
security.OCSP.enabled is an integer


Genbox:
security.OCSP.enabled is set to true in your config, however, it is actually an integer. 1 is default in Firefox 3 and above.

pyllyukko:
Merged

Issue #29, opened 2015-08-08
Spelling mistake


Genbox:
Is DuckDcukGo not supposed to be DuckDuckGo?

user_pref("browser.search.defaultenginename",			"DuckDcukGo");

pyllyukko:
Merged

PR #35, opened 2015-08-09
Update user.js to remove duplicates, invalid entries and spelling mistakes

CHEF-KOCH:
* fix tab spacing (two tabs)
* removed chiper suite duplicates  #32
* fixed security.OCSP.enabled #30
* Fixed spelling mistake in the default search engine #29

Special thanks goes to Genbox to pointing this out.

### DIFF
--- a/user.js
+++ b/user.js
@@ -10,46 +10,46 @@
 
 // disable Location-Aware Browsing
 // http://www.mozilla.org/en-US/firefox/geolocation/
-user_pref("geo.enabled",			false);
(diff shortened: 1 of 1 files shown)


pyllyukko:
I would still greatly appreciate, if you could break these pull requests into several commits. Each issue in it's own commit at least.

CHEF-KOCH:
Well, I try my best but it's more work for me, since I need to split my local commits and merge this with yours. I hope my git client gets an update soon to solve this.

PR #39, opened 2015-08-16
Update README.md

CHEF-KOCH:
Better todo look 
Make all links click able and tag them with a proper title
Comparison: shorten the HTML5 test link, broke my mobile browser and I need 2 minutes to scroll to the end because small screen :expressionless:

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,14 +244,14 @@ Online tests
   * [Similar from Microsoft](https://ie.microsoft.com/testdrive/browser/mixedcontent/assets/woodgrove.htm)
 * [WebRTC stuff](http://mozilla.github.io/webrtc-landing/)
 * [Flash player version](https://www.adobe.com/software/flash/about/) from Adobe
-* [Verify Java Version](https://www.java.com/en/download/installed.jsp)
-  * Protip: Don't use Java!! But if you do, at least have it updated.
+* [Verify your installed Java Version](https://www.java.com/en/download/installed.jsp)
+  * Protip: Don't use Oracle's Java!! But if you really need it, update it regulary!
 * [IP check](http://ip-check.info/?lang=en)
-* http://cure53.de/leak/onion.php
+* [Onion test for CORS and WebSocket](http://cure53.de/leak/onion.php)
 * [Firefox Addon Detector](http://thehackerblog.com/addon_scanner/)
   * [Blog post](http://thehackerblog.com/dirty-browser-enumeration-tricks-using-chrome-and-about-to-detect-firefox-plugins/)
 * [browserrecon](http://www.computec.ch/projekte/browserrecon/)??
-* http://get.webgl.org/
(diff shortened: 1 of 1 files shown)


CHEF-KOCH:
Btw: About the Java note, the software isn't that vulnerable anymore since they minimum update them once in a month and on critical stuff asap. So, we maybe need to talk about this, in generally the java language isn't vulnerable, it's just the software/installer.

Hunter-Github:
Leaving Java on increases your attack surface and there are very few decent sites that use Java applets anyway.

CHEF-KOCH:
Java loads only if something requests it, so the attack scenario are mostly on a 0day level. The normal security problems are fixed very fast (2008: 4832 security problems - compared to 2014 ~100 known 'thats almost nothing'), the chance to browse on an vulnerable page + that it use a 0day or an attack that still works on your Java version + breaks the sandbox + you enabled it via click-to-play is more or less theoretically. It's a lot of easier to break into the system itself by bypassing e.g. the firewall with NTLM-Hash (alias 'pass-the-hash') attacks [that never gets fixed btw!]. So the whole Java security problems discussion is FUD without everything except that it was previously maybe more vulnerable which is an fact. 

Thanks for the link, will look at this.

pyllyukko:
> Leaving Java on increases your attack surface and there are very few decent sites that use Java applets anyway. Relevant: https://security.stackexchange.com/a/19603

I agree. It's still quite valid attack vector and it isn't very common for websites to have Java applets anymore (luckily).

CHEF-KOCH:
Thanks for merging this.

About Java:
I disagree, even after reading the given link, this only shows that an attacker could use special api's too bypass xyz - bu that's the nature of almost every language. Without that dangerous api's some stuff would never work, because it is there doesn't mean it will be used. And again, the attack scenario is very low, if not impossible. Imho JavaScript and Flash is much more vulnerable these days. Most Tor attacks on a JavaScript level, and remember that Tor allows JavaScript (in NoScript) by default, if not this would break most of all pages and there is no global whitelist avaible which shows which sites are secure (btw even secure pages could be hacked!). Please don't forgot that Java doesn't comes pre-installed on most of the systems (speaking about the installer) so we again talking about things that are user controlled, saying it's insecure without everything could applied to every other software you installed too. 

* http://www.oracle.com/technetwork/java/seccodeguide-139067.html
* https://en.wikipedia.org/wiki/Criticism_of_Java
* www.aabri.com/manuscripts/131731.pdf
* http://www.csoonline.com/article/2875535/application-security/java-is-the-biggest-vulnerability-for-us-computers.html
* https://www.whitehatsec.com/assets/WPstats_spring10_9th.pdf

So to say this is more vulnerable or more attackable compared to other things is pure wrong. With such arguments you better do not use any Browser, in fact they come with the same risks, see also latest fingerprint leaks and pdf exploits.

PR #40, opened 2015-08-19
Fix borked network resource authentication credentials dialog on ≤ 39

CHEF-KOCH:
Since FF 40/40.0.2 the sub-resources HTTP-authentication seems to be incorrect, so we should prefer an integer of 2.

Sub-resources HTTP-authentication:
0 - don't allow sub-resources to open HTTP authentication credentials dialogs
1 - allow sub-resources to open HTTP authentication credentials dialogs, but don't allow it for cross-origin sub-resources
2 - allow the cross-origin authentication as well

user_pref("network.auth.allow-subresource-auth", 2);

### DIFF
--- a/user.js
+++ b/user.js
@@ -8,13 +8,13 @@
  *                                                                            *
  ******************************************************************************/
 
-// disable Location-Aware Browsing
+// Disable Location-Aware Browsing
 // http://www.mozilla.org/en-US/firefox/geolocation/
 user_pref("geo.enabled",		false);
 
 // http://kb.mozillazine.org/Dom.storage.enabled
 // http://dev.w3.org/html5/webstorage/#dom-localstorage
-// you can also see this with Panopticlick's "DOM localStorage"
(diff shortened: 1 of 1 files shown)


pyllyukko:
Seems like setting this to ```0``` would be the most secure option. Why ```2```?

CHEF-KOCH:
Because it is set to 0 by default which course the problem. 1 might work but possible break addons like KeeFox (and I use this).

pyllyukko:
Setting it to ```2``` makes the browser vulnerable to attacks, so we really can't have this. I think we might even want to change it to ```0``` just to be on the safe side.

CHEF-KOCH:
There is not even one single CVE or any theoretically attack (POC) for it but okay. There are problems with stronger settings, see [here](http://forums.mozillazine.org/viewtopic.php?f=38&t=2955671). It also could break some RSS feeds which may need additional login to get them.

Issue #33, opened 2015-08-09
Consider removing old settings


Genbox:
To simplify the config, consider removing old settings that no longer apply. As of Firefox 39.0.3, the following settings no longer apply:

- [x] browser.frames.enabled
- [ ] browser.download.manager.retention
- [ ] browser.history_expire_days
- [ ] browser.history_expire_sites
- [ ] browser.history_expire_visits
- [ ] general.useragent.override
- [ ] plugins.hide_infobar_for_outdated_plugin

Hunter-Github:
Would disagree with that: there are also ESR browsers out there. Historically significant options may be simply put in a single section of the file.

Genbox:
This issue is just a consideration, not an actual issue which needs to be fixed right away. I think it would be prudent to keep track of deprecated settings along the way, and remove them as new releases of Firefox ESR comes out.

That being said, The the settings are also not present in the current version of Firefox ESR.

Genbox:
The following settings are also not present in current Firefox and Firefox ESR releases.

browser.sessionstore.enabled
browser.sessionstore.postdata
javascript.options.methodjit.chrome
javascript.options.methodjit.content
media.gmp-gmpopenh264.enabled
network.dns.disablePrefetchFromHTTPS
spdy.enabled.v3
network.negotiate-auth.allow-insecure-ntlm-v1
network.negotiate-auth.allow-insecure-ntlm-v1-https
security.enable_ssl3
security.enable_tls_session_tickets
security.ssl3.dhe_rsa_camellia_128_sha
security.ssl3.dhe_rsa_camellia_256_sha
security.ssl3.dhe_dss_camellia_128_sha
security.ssl3.dhe_dss_camellia_256_sha
security.ssl3.rsa_camellia_128_sha
security.ssl3.rsa_camellia_256_sha
security.ssl3.rsa_null_md5
security.ssl3.rsa_null_sha
security.ssl3.ecdh_rsa_null_sha
security.ssl3.ecdhe_ecdsa_null_sha
security.ssl3.ecdhe_rsa_null_sha
security.ssl3.ecdh_ecdsa_null_sha
security.ssl3.dhe_dss_aes_128_sha
security.ssl3.dhe_dss_aes_256_sha
security.ssl3.dhe_dss_des_ede3_sha
security.ssl3.dhe_rsa_des_ede3_sha
security.ssl3.ecdh_ecdsa_aes_128_sha
[...]

PR #42, opened 2015-08-20
Separate old settings in his own category

CHEF-KOCH:
Merged #33
The ciphers are not comment out, all the rest can be changed as you want it 
general.useragent.override still works
small other changes

### DIFF
--- a/user.js
+++ b/user.js
@@ -4,6 +4,108 @@
  ******************************************************************************/
 
 /******************************************************************************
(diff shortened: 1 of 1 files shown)


szepeviktor:
+1

pyllyukko:
These kind of major structure changes need to be thought out more, closely investigated setting by setting and performed with many many commits, so that's it clear on the log how and what has changed.

Issue #13, opened 2015-04-30
Update for Firefox [currently: 47.0.0]


CHEF-KOCH:
Hey, thanks, the project seems promising, it's exactly what I'm looking for, but sad it's pretty outdated atm. Some settings are deprecated (means the value is already set to the ones mentioned in the user.js) and some not exists anymore due security reasons (e.g. most of the cipher stuff).

Here is my list what I've found, I checked each entry against the current Firefox release.

Not existing anymore:

``` bash
browser.frames.enabled 
browser.polaris.enabled
javascript.options.methodjit.chrome
javascript.options.methodjit.content
browser.safebrowsing.remoteLookups
datareporting.healthreport.uploadEnabled
browser.newtabpage.enhanced
network.negotiate-auth.allow-insecure-ntlm-v1
browser.sessionstore.postdata
browser.sessionstore.enabled
browser.history_expire_days
browser.history_expire_sites
browser.history_expire_visits
browser.download.manager.retention
plugins.hide_infobar_for_outdated_plugin
security.warn_viewing_mixed
security.warn_entering_weak
network.http.spdy.enabled.v3
security.enable_tls_session_tickets
security.enable_ssl3
security.ssl3.rsa_null_sha
security.ssl3.rsa_null_md5
security.ssl3.ecdhe_rsa_null_sha
security.ssl3.ecdhe_ecdsa_null_sha
security.ssl3.ecdh_rsa_null_sha
security.ssl3.ecdh_ecdsa_null_sha
security.ssl3.rsa_seed_sha
security.ssl3.rsa_rc4_40_md5
security.ssl3.rsa_rc2_40_md5
security.ssl3.rsa_1024_rc4_56_sha
security.ssl3.rsa_camellia_128_sha
security.ssl3.ecdh_rsa_rc4_128_sha
security.ssl3.ecdh_rsa_aes_128_sha
security.ssl3.ecdh_ecdsa_rc4_128_sha
security.ssl3.ecdh_ecdsa_aes_128_sha
security.ssl3.dhe_rsa_camellia_128_sh
security.ssl3.dhe_dss_des_ede3_sha
security.ssl3.dhe_rsa_des_ede3_sha
security.ssl3.ecdh_ecdsa_des_ede3_sha
security.ssl3.ecdh_rsa_des_ede3_sha
security.ssl3.ecdhe_ecdsa_des_ede3_sha
security.ssl3.ecdhe_rsa_des_ede3_sha
[...]

PR #44, opened 2015-08-23
NoScript NSA will not work anymore after FF42+

CHEF-KOCH:
added a warning about coming changes, for more details take a look [here](https://github.com/pyllyukko/user.js#issuecomment-130116288)

The addon itself is okay, but Mozilla seems to massive improve the entire add-ons stuff, after FF42 this can't be reversed with any about:config tweak (in the name of security) other add-ons are maybe also affected by this e.g. HTTPSE since this isn't in the store and also use old API's that getting removed, we will see if there will be an update or not, in the future HTTPSE will be obsolete anyway since this is already checked with FF 41+.

Only ESR versions are not affected by this (atm).

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ See also:
 
 #### Add-ons for mobile platforms
 
-* [NoScript Anywhere](https://noscript.net/nsa/)
+* [NoScript Anywhere](https://noscript.net/nsa/) - will not work anymore after FireFox 42+, [due addons and signature changes](https://github.com/pyllyukko/user.js#issuecomment-130116288)
 * [uBlock](https://addons.mozilla.org/en-US/android/addon/ublock-origin/)
 * [HTTPS Everywhere](https://www.eff.org/https-everywhere)
 


pyllyukko:
The link doesn't seem to be correct.

CHEF-KOCH:
Strange the link is correct but GitHub seems to have problems (maybe they changed the API again). I'm talking about my post over #13 (FF 40 changes) the first link offers that in the near future Chrome and Opera addons and possible plugins will work with Firefox too, because of this changes some deprecated API getting removed + the fact that NSA NoScript isn't available on the store and not signed (yet). I already tried to contact the developer and webmaster of this project (without luck).

Here is the link only:
https://blog.mozilla.org/addons/2015/08/21/the-future-of-developing-firefo

CHEF-KOCH:
There is an solution, it's called [OrFox](https://github.com/guardianproject/orfoxfennec) and the successor of OrWeb. Technical speaking it's a Firefox based Browser with an harden configuration for [Tor](https://www.torproject.org/projects/torbrowser/design/). It includes HTTPS-Everywhere and NoScript (like in the normal Tor Browser Bundle). The Browser needs Orbot installed and can be installed via this link e.g. via F-Droid <code>https://dev.guardianproject.info/debug/info.guardianproject.orfox/fdroid/repo</code>.

CHEF-KOCH:
Well I found a _solution_, NSA will still work, [look here for the patch](https://forums.informaction.com/viewtopic.php?f=25&t=21189&sid=c93da8c3d7cf8cbbfc6aa9fbb7200907).

General this is the FF history
* Firefox 40 - Warnings are shown if unsigned extensions are installed.
* Firefox 41 - Warnings continue to show up.
* Firefox 42 - Warnings are still displayed if unsigned add-ons are run.
* Firefox 43 - Add-ons without signatures are blocked by default, but there is an override that is available in all versions of the browser.

The basic question is if it's better to drop and use OrFox or make it compatible with patches.

I will report back if there is an [official statement](https://forums.informaction.com/viewforum.php?f=25). At the moment I can't recommend it because it's not available at the store + the fact that there are only 3 maintainer. 
Firefox 44 - Only signed add-ons can be installed in Firefox Stable and Beta. There is no override anymore for those editions of the browser. The override remains in Firefox Developer and Nightly

pyllyukko:
If you fix the link in another commit, I can merge this?

CHEF-KOCH:
Yep, I will bump some new stuff on weekend. As from what I know, according to some other forum posts if a user installed an old (non patched) version and not deactivated the FF internal update mechanism, he will get the .x10 version automatically.

szepeviktor:
`https://github.com/pyllyukko/user.js#issuecomment-130116288`
The issue number is missing from this URL.

CHEF-KOCH:
Latest build can be found [here](https://noscript.net/nsa/latest), I close it since it's solved.

PR #48, opened 2015-09-23
Bigger update to fit with HTTPSE and NoScript

CHEF-KOCH:
Since we're recommend (and using) NoScript and HTTPS Everywhere, we can just harden some more stuff from the beginning. 

HTTPS Everywhere:
* Enable all popup messages to notify on observatory problems/changes/news
* Enable the toolbar hint (not annoying)

NoScript:
The settings are to fit with Firefox, so e.g. we doesn't want to allow webgl (except on whitelistening pages the user sets - if not set also apply these restrictions to whitelist sites too -> must be manually set because it's up to the user).

* Enable ABE (don't do this on Tor)
* Enable STS (don't do this on Tor)
* Don't allow the cliuck-to-play stuff, it's depcreated and only for older FF versions (or maybe ESR variants of it) 
* Only allow the current tab on change (not all opened tabs)
* Cascade the menu, it's a visual tweak 
* Show some other visual options to not get useless things displayed and reveal only stuff that is important, uBlock or uMatrix already do a better finer job for this 
* Changed the default policy to harden possible security related 'leaks' 
* Media is enabled since this breaks so many pages (due .js, HTML5) [use Random Agent Spoofer or uMarix for this]

### DIFF
--- a/user.js
+++ b/user.js
@@ -619,3 +619,62 @@ user_pref("security.ssl3.dhe_dss_camellia_256_sha",		false);
 // Fallbacks due compatibility reasons
 user_pref("security.ssl3.rsa_aes_256_sha",		true);
 user_pref("security.ssl3.rsa_aes_128_sha",		true);
+
+/******************************************************************************
+ * HTTPS Everywhere Preferences:                                              *
+ * Harden some stuff that isn't visible by default                            *
+ ******************************************************************************/
+user_pref("extensions.https_everywhere._observatory.popup_shown", true);
+user_pref("extensions.https_everywhere.toolbar_hint_shown", true);
+
+
+/******************************************************************************
+ * NoScript:                                                                  *
+ * A compromise between security and usability                                *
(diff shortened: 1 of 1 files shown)


user.js
```diff
@@ -619,3 +619,62 @@ user_pref("security.ssl3.dhe_dss_camellia_256_sha",		false);
...
+user_pref("extensions.https_everywhere.toolbar_hint_shown", true);
+
+
+/******************************************************************************
+ * NoScript:                                                                  *
+ * A compromise between security and usability                                *
+ ******************************************************************************/
+
+
+pref("capability.policy.maonoscript.javascript.enabled", "allAccess");
```
pyllyukko: Why it's ```pref()``` instead of ```user_pref()```?

pyllyukko:
I don't think we should have settings for extensions. Let's just keep this user.js for the core Firefox stuff.

PR #56, opened 2015-10-09
SPDY will replaced by http2.0

CHEF-KOCH:
This entries have no longer affect and making troubles on some nxig servers because http2 and spdy aren't compatible.

### DIFF
--- a/user.js
+++ b/user.js
@@ -471,12 +471,6 @@ user_pref("security.ask_for_password",		0);
 // https://wiki.mozilla.org/Privacy/Features/HSTS_Preload_List
 user_pref("network.stricttransportsecurity.preloadlist",		true);
 
-// enable SPDY
-// https://en.wikipedia.org/wiki/SPDY
-user_pref("network.http.spdy.enabled",		true);
-user_pref("network.http.spdy.enabled.v3",		true);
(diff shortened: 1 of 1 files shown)


pyllyukko:
Any links/references/documentation relating to this issue?

CHEF-KOCH:
Issue? Huh, this is directly from the foundation. Some pages using SPDY but the option will be removed + it isn't compatible with the final HTTP 2.0 (because SPDY was just only a test). 

It's recommadation from the foundation that's why I bumped it. If you are on the SPDY page it will override your browser settings.

Reference:
* https://en.wikipedia.org/wiki/HTTP/2 (see 'HTTP/HTTPS servers' + 'Genesis in and later differences from SPDY'
* https://github.com/http2/http2-spec (official specs)
* https://http2.github.io/faq/ FAQ
* http://blog.chromium.org/2015/02/hello-http2-goodbye-spdy-http-is_9.html

And about the NGIX Statement, it was on there Homepage, can't find it now but the draft also confirms it is not SPDY compatible, so HTTP2 will be used or in some cases backwards to SPDY but not both. Killing the options also not kills the ability to visit the pages (tested with FF 41.0.2). 
*

PR #57, opened 2015-10-13
added JavaScript exploit protection (optional)

CHEF-KOCH:
Optional protection against eval() to execute data like it was code.

Why optional?
it possible breaks some legit site (normal sites should not use eval but all is possible and it must be exluded manually)

### DIFF
--- a/user.js
+++ b/user.js
@@ -8,6 +8,13 @@
  *                                                                            *
  ******************************************************************************/
 
+// Protection against Javascript exploits to block eval() 
+// See: https://hackademix.net/2011/09/29/script-surrogates-quick-reference/
+//noscript.surrogate.noeval.replacement,		window.eval = null;document.eval=null;
+//noscript.surrogate.noeval.sources,		@^http://[a-z]+[^/]+\.[a-z]+(?:/|$);
+//noscript.surrogate.noeval.exceptions;
+
+
 // disable Location-Aware Browsing
 // http://www.mozilla.org/en-US/firefox/geolocation/
 user_pref("geo.enabled",		false);


berrythesoftwarecodeprogrammar:
make the source @^https?:\/\/[a-z]+[^/]+\.[a-z]+(?:\/|$)
(idk if all the escapes are necessary but should support https)

CHEF-KOCH:
It's not necessary since 1) https pages are secure (and I also hope the cookies, if not use the cookie option within NoScript) and 2) we use HTTPS-Everywhere which automatically redirects to https.

berrythesoftwarecodeprogrammar:
hmm how do you mean? a https site could still use eval the same way as non https sites right? just wondering, im not sure

CHEF-KOCH:
There is no site I know which uses https which suffering from this attack. Normally sites which using this aren't secure because e.g. there is no reason to secure html5 validator/editor pages an example page seems [this](http://www.w3schools.com/).

If you found any site I will add it. But from what I know a real secure pages never use such strange .js stuff.

berrythesoftwarecodeprogrammar:
within a few minutes of disable eval, i noticed 3 sites break. some of facebook.com's functionality. google images on encrypted.google.com. and wolframalpha.com. it seems like more sites use it than originally thought. i still like being able to disable it however.
currently using:
```
noscript.surrogate.noeval.sources, @^https?\:\/\/.*$
noscript.surrogate.noeval.exceptions, ^.*(?:facebook\.com|encrypted\.google\.com|wolframalpha\.com).*$
```
(because it seems like exceptions doesnt support a list, just one regex)

CHEF-KOCH:
Yeah, because of such problems that's the reason why it's optional. Or just add them to exclusions list as you already did now.

Your mentioned sites are bad examples, they often change there api's, means that your statement will possible next week already been wrong. Such pages shouldn't been blocked or just totally (if you not need them).

Maybe random agent spoofer is better for this, but in his currently state it also suffers from the same problem, the internal whitelist must be updated. 

However, after some talk I asked if the safe-browsing mechanism could block this anyway without breaking anything (because if, I'm sure it will be immediately fixed).

pyllyukko:
There's enough to go around with only Firefox's settings, so I'm reluctant about starting to tweak various add-ons also.

Also, disallowing ```eval``` doesn't make much difference and as @berrythesoftwarecodeprogrammar said, it breaks sites.

PR #66, opened 2015-10-21
Explicitly enable subresource integrity

fmarier:
Like CSP, it's enabled by default, but if we're including CSP,
we should probably include the pref for SRI too.


CHEF-KOCH:
This option doesn't exist anymore.

PR #67, opened 2015-10-24
Disabled SSL cert. error reporting

CHEF-KOCH:
Disabled the SSL error reporting stuff, better use HTTPSE 
https://gecko.readthedocs.org/en/latest/browser/base/sslerrorreport/preferences.html

Optional because you maybe not use HTTPSE Observatory or your AV doesn't comes with it's own reporting feature (or you simply use no AV).

### DIFF
--- a/user.js
+++ b/user.js
@@ -484,6 +484,10 @@ user_pref("security.ask_for_password",		0);
  *                                                                            *
  ******************************************************************************/
 
+// Disable SSL Certification error reporting, since we use HTTPSE we not need both
+//See: https://gecko.readthedocs.org/en/latest/browser/base/sslerrorreport/preferences.html
+//user_pref("security.ssl.errorReporting.enabled,		false);
+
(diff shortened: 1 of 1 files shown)


user.js
```diff
@@ -484,6 +484,10 @@ user_pref("security.ask_for_password",		0);
  *                                                                            *
  ******************************************************************************/
 
+// Disable SSL Certification error reporting, since we use HTTPSE we not need both
+//See: https://gecko.readthedocs.org/en/latest/browser/base/sslerrorreport/preferences.html
+//user_pref("security.ssl.errorReporting.enabled,		false);
```
pyllyukko: There's a double quote missing, which breaks the JavaScript syntax.

pyllyukko:
I think this is not necessary. See the reasoning on 263f5b26cc106d1d8bdd2f57512a675a53f8b1a3.

CHEF-KOCH:
Yupp that's true didn't saw it. Thanks.

PR #69, opened 2015-10-25
Preventing css history leakage

CHEF-KOCH:
https://blog.mozilla.org/security/2010/03/31/plugging-the-css-history-leak/


pyllyukko:
So this is just spelling fixes? If so, you should state that in the commit message.

pyllyukko:
We already have this: https://github.com/pyllyukko/user.js/blob/master/user.js#L459-L461

PR #70, opened 2015-10-25
added mozilla security source research

CHEF-KOCH:
* added research

This seems promising, because the projects in it like the openh264 stuff may can be adopted in the future, it's also always useful to get some more research, especially because here is all the necessary source.


pyllyukko:
This doesn't relate that much to Firefox itself.

PR #71, opened 2015-10-25
added update hardening

CHEF-KOCH:
Biggest weakness I can see it still the about:config update change problem in FF so we ensure we use all the latest url's. Of course if you decided to no use e.g. peerconnection it makes less sense but not everyone disabled it so give them the ability to use the latest secured servers and update resources.

### DIFF
--- a/user.js
+++ b/user.js
@@ -4,6 +4,28 @@
  ******************************************************************************/
 
 /******************************************************************************
+ * Update hardening                                                           *
+ * Since FF never updates url's i about:config we ensure we get the latest    *
+ * software, updates and ressources.                                          *
+ ******************************************************************************/
+
+user_pref("app.update.url",		"https://aus5.mozilla.org/update/3/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/%DISTRIBUTION%/%DISTRIBUTION_VERSION%/update.xml");
+user_pref("browser.trackingprotection.gethashURL",		"https://shavar.services.mozilla.com/gethash?client=SAFEBROWSING_ID&appver=%VERSION%&pver=2.2");
+user_pref("browser.trackingprotection.updateURL",		"https://shavar.services.mozilla.com/downloads?client=SAFEBROWSING_ID&appver=%VERSION%&pver=2.2");
+user_pref("media.gmp-manager.certs.1.commonName",		"aus5.mozilla.org");
+user_pref("media.gmp-manager.certs.1.issuerName",		"CN=DigiCert SHA2 Secure Server CA,O=DigiCert Inc,C=US");
+user_pref("media.gmp-manager.certs.2.commonName",		"aus5.mozilla.org");
+user_pref("media.gmp-manager.certs.2.issuerName",		"CN=thawte SSL CA - G2,O="thawte, Inc.",C=US");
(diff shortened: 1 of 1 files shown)


fmarier:
I don't get why you're explicitly setting these URLs to the defaults. If you don't customize a pref and leave it to be the default, Firefox will update it automatically if it changes in the future.

CHEF-KOCH:
No, not from my experience, FF never ovverides on updates (dirty install over existent installation) such kind of url's. I tried it, e.g. use FF 39 and start it, close it and upgrade to a newer one and you will see it still uses the old servers.

fmarier:
> No, not from my experience, FF never overrides on updates (dirty install over existent installation) such kind of url's. I tried it, e.g. use FF 39 and start it, close it and upgrade to a newer one and you will see it still uses the old servers.

If you can reproduce this, please file a bug against Firefox. It would be a fairly important bug.

CHEF-KOCH:
I can reproduce this, but as I said over another thread I lost my entire BugZilla account. I think the entire about:config must be re-written because there are several more 'problems/bugs' or 'by design' thins that are really annoying, it start's with the fact that there is no API to detect and automatically revert changes (must use an new profile). Most plugins reverting this with an additional .js which compares current settings with there provided one. :-1:

nodiscc:
I can't reproduce this on Iceweasel 41.0.2-1

CHEF-KOCH:
Did you create another profile for test? Because in such a case Mozilla always starts with defaults (and of course with the new version with it's new defaults). Mozilla changed that now long time ago, try to upgrade form an older version and you will see.

* Eg use FF 33 
* Download and dirty install FF 41 over it 
* Now you will see

fmarier:
Download [Firefox 38](https://www.mozilla.org/en-US/firefox/organizations/all/) and create a new profile:

    ./firefox --no-remote -P

Then open `about:config` and notice this:

    browser.trackingprotection.gethashURL = "https://tracking.services.mozilla.com/gethash?client=SAFEBROWSING_ID&appver=%VERSION%&pver=2.2"

Close Firefox.

Now download [Firefox 42](https://www.mozilla.org/en-US/firefox/beta/all/) and open the same profile.

Look at the same pref in `about:config`:

    browser.trackingprotection.gethashURL = https://shavar.services.mozilla.com/gethash?client=SAFEBROWSING_ID&appver=%VERSION%&pver=2.2

The URL was automatically updated in the profile **because** it hadn't been customized. So merging this PR will actually prevent Firefox from keeping URLs up to date.

CHEF-KOCH:
Huh, I use shavar.services.mozilla.com in my commit.

But in this case we still have the issue that such important parts not getting updated so how to handle?

nodiscc:
Such parts _do_get updated _if_ you left the default value intact. The solution is to not alter these values at all.

pyllyukko:
I guess that settles it.

CHEF-KOCH:
I can't confirm with FF 33 and default settings, but okay. Maybe this was/is an internal bug but I still think that this is useful anyway to stay on the line because if someone uses an older version he also want's to get the latest update resources. So and if he not upgrades or created an new profile he will suffer from this. 

So this is for me not understandable to nor merge it. But okay.

pyllyukko:
There's already plenty of work trying to keep up with all the settings and features, so I think in that sense also, it's best to let FF handle these.

PR #72, opened 2015-11-03
add amiunique and canvasblocker doc

nodiscc:
add resources for future investigation:

 * amiunique (open source browser fingerprinting tool)
 * canvas-blocker firefox addon related to canvas fingerprinting research

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,6 +256,8 @@ Online tests
 * [battery.js](https://pstadler.sh/battery.js/)
 * [RC4 fallback test](https://rc4.io/)
 * [Battery API](https://robnyman.github.io/battery/)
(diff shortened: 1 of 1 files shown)


CHEF-KOCH:
Well I still don't get it why we need to add every theoreically PoC to show what could be. I never saw any of these attacks yet on 'trusted' pages. As I already said, it's like downloading malware to show (if you disable all OS protection) what could it do. Makes not much sense, wouldn't it be better to solve daily security problems?

I already wrote about canvas blocker in another thread, so why not talk about it in the already existent thread? If RAS gets an update it can handle it too with additional things, so instead of using this I still think random agent spoofer combines more and even without the two things noScript already can block canvas.

PR #76, opened 2015-11-19
Make private browsing optional

CHEF-KOCH:
Removed private browsing and mark it as optional since there is no benefit except to break addons

The test example pages doesn't care about private browsing mode or not, Mozilla not enabled by default for several reasons so I think without any real benefit we are already safe to go with NoScript + uBlock + some Options which are already adopted to FF (e.g. clean all Cookies at shutdown and such).

### DIFF
--- a/user.js
+++ b/user.js
@@ -359,8 +359,8 @@ user_pref("browser.cache.offline.enable",		false);
 // Always use private browsing
 // https://support.mozilla.org/en-US/kb/Private-Browsing
 // https://wiki.mozilla.org/PrivateBrowsing
-user_pref("browser.privatebrowsing.autostart",		true);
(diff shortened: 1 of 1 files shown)


pyllyukko:
The benefit is not having (potentially) sensitive information stored on your disk. It's not all about websites tracking you, but also covering your own ass in case of someone getting ahold of your data.

Defence in depth that is.

CHEF-KOCH:
Sorry I not understand your explanation. 

For me it breaks specific addons which are designed to 'secure' your browser experience and from what I know NoScript already did the job (even in private/non-private mode). So what exactly is the benefit. This was only a reason to use while the normal mode couldn't clean hsts but this was improved in NoScript and with never Firefox releases. 

Maybe I miss something here.

PR #77, opened 2015-11-19
Whoer added which also detects Flash IP leakage

CHEF-KOCH:
^^


pyllyukko:
People shouldn't be running Flash in the first place :) I think we have enough online tests already, if this site doesn't provide anything else unique.

PR #78, opened 2015-11-19
Offline test section added

CHEF-KOCH:
First tool added, which checks Windows server based systems. It comes with several features and compared to gpedit.msc it's better since this writes things directly into registry which doesn't have any char limit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,6 +259,14 @@ Online tests
 * [AmIUnique](https://amiunique.org/) ([Source](https://github.com/DIVERSIFY-project/amiunique))
 
 
+
+Offline tests
+-------------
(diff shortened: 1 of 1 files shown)


nodiscc:
Hi, this is a server oriented tool, am I right? If so I don't know if it it really belongs here. There are tons of tools and documentation on how to configure servers more securely (thinking about cipherli.st, apache2-modsecurity, mod-evasive etc..) but you may not want to include all of this  in the user.js doc, which is after all, about securing Firefox.

pyllyukko:
Yup. We're not hardening servers on this project. Especially IIS ;)

PR #79, opened 2015-11-19
Disable the ability for Firefox to report what plugins are installed

CHEF-KOCH:
Block websites from intentionally removing content when they detect certain plugins e.g like Facebook. This is optional since it will break the entire plugin check page from Mozilla but it helps to avoid been fingerprinted.

### DIFF
--- a/user.js
+++ b/user.js
@@ -17,9 +17,11 @@ user_pref("geo.enabled",		false);
 // you can also see this with Panopticlick's "DOM localStorage"
 //user_pref("dom.storage.enabled",		false);
 
-// Don't reveal internal IPs
+// Don't reveal internal IPs and reduce fingerprinting
 // http://net.ipcalf.com/
 user_pref("media.peerconnection.enabled",		false);
+//user_pref("plugin.scan.plid.all", false);
+
 // getUserMedia
 // https://wiki.mozilla.org/Media/getUserMedia
 // https://developer.mozilla.org/en-US/docs/Web/API/Navigator


nodiscc:
> Block websites from intentionally removing content when they detect certain plugins

Nice feature/anti-feature, didn't know it existed. Is this the feature that panopticlick uses to detect installed plugins?

You may want to link to http://kb.mozillazine.org/About:config_entries#Plugin. and/or http://kb.mozillazine.org/Plugin_scanning in the comments, and note that this seems to only affect Windows installs.

pyllyukko:
From what I understand from <http://kb.mozillazine.org/Plugin_scanning>... it sounds more like Firefox (and other Mozilla software) scanning for installed/available plugins on the host and not something that happens from a website.

CHEF-KOCH:
>  Is this the feature that panopticlick uses to detect installed plugins?

From what I know it uses name enumeration (plugins.enumerable_names removed since FF41) + this to detect your plugins/configuration. This feature is registry based scanning. Enabling this also disabled the e.g. Flash ability. I don't think this is only Local, since some websites checking your registry to determinate if you e.g. use Flash. 

```
x86
HKEY_CURRENT_USER\Software\MozillaPlugins\plugin-id
HKEY_LOCAL_MACHINE\Software\MozillaPlugins\plugin-id

64-bit Windows:
HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\MozillaPlugins\plugin-id
```

pyllyukko:
I'm still not sure whether this setting matters that much. If I understood correctly, it's the browser that is scanning available plugins from the OS and not a website enumerating the browser.

nodiscc:
Right. Let's close this? Poorly documented

pyllyukko:
Let's.

PR #80, opened 2015-11-19
added 'How can I whitelist specific websites'

CHEF-KOCH:
Added a whitelist option without the necessary of needed to install any software/addon, it's quick and dirty but it works.


pyllyukko:
So you mean block-everything-else-but-these kinda whitelist?

CHEF-KOCH:
Yep.

pyllyukko:
I think this is a bit too "hacky" to be recommended.

PR #82, opened 2015-11-20
Disabled face detection

CHEF-KOCH:
^^

### DIFF
--- a/user.js
+++ b/user.js
@@ -85,6 +85,9 @@ user_pref("webgl.disabled",		true);
  *                                                                            *
  ******************************************************************************/
 
+// Disable face detection by default
+user_pref("camera.control.face_detection.enabled";    false);
+
 // Default search engine
 //user_pref("browser.search.defaultenginename",		"DuckDuckGo");
 


pyllyukko:
Any links / references?

CHEF-KOCH:
Nah, not really, I found it myself.

But there is:
* https://gist.github.com/haasn/69e19fc2fe0e25f3cff5
* https://malwaretips.com/threads/firefox-about-config-privacy-security.50957/

I found no other switch which face detection so this must be the only one, and even if you per-site control the camera option, there is no face detection switch/toggle. Same like the VR option but the difference here is that VR is disabled by default while e.g. the oculus specific setting is enabled.

pyllyukko:
There was a syntax error in your commit. I fixed it in c6b1c0c6d3a0d8111f1264a97cc56b93ba289903, but please be careful regarding these in the future? Thanks.

CHEF-KOCH:
Sr, such things can be happen if you not copy & paste all things. I will check in the future more carefully.

PR #84, opened 2015-11-22
No need to completely disable WebRTC anymore

CHEF-KOCH:
No need to disable WebRTC anymore, because WebRTC is not overall bad and some pages will be not working anymore if you disabled it competently. Now we got an option to use it without leaking the IP, because now only the IP will be used that is visible to the server.

### DIFF
--- a/user.js
+++ b/user.js
@@ -17,10 +17,12 @@ user_pref("geo.enabled",		false);
 // you can also see this with Panopticlick's "DOM localStorage"
 //user_pref("dom.storage.enabled",		false);
 
-// Don't reveal internal IPs
-// http://net.ipcalf.com/
+// Don't reveal your internal IP
+// Check the settings with: http://net.ipcalf.com/
 // https://wiki.mozilla.org/Media/WebRTC/Privacy
 user_pref("media-peerconnection.ice.default_address_only",		true);
+//user_pref("media.peerconnection.enabled",		false);
+
 // getUserMedia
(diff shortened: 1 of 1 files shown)


pyllyukko:
Nice! We still need to wait a bit, as Firefox 42 still reveals this info. And on top of that some grace period, so users have actually updated Firefox.

pyllyukko:
Can you modify this, so that ```media.peerconnection.enabled``` is still disabled (left as is) and the new setting is added? And maybe have a separate issue/pull request for the ```media.peerconnection.enabled```. I think we should still leave the ```media.peerconnection.enabled``` into the user.js, but commented out.

CHEF-KOCH:
Is my commit now okay :+1: or :-1: ?

pyllyukko:
I meant that leave the ```media.peerconnection.enabled``` as it was.

CHEF-KOCH:
Update, it's now implemented since [46.0 Beta 9](https://wiki.mozilla.org/Media/WebRTC/ReleaseNotes/46).

nodiscc:
Firefox 52 is the new [ESR](https://www.mozilla.org/en-US/firefox/organizations/all/). So we should either:

 * comment out the pref disabling WebRTC (`media.peerconnection.enabled`)
 * leave it as is for hardening purposes, and close this issue
  * if notable functionality breakage is found due to WebRTC being disabled, add a `NOTICE:` field alongside that pref.

PR #85, opened 2015-11-24
added Certificate Checks

CHEF-KOCH:
This section is special, this is for known vulnerable certificates. See the Dell story, Superfish and [more](http://www.ghacks.net/2015/11/23/dell-does-a-lenovo-ships-laptops-with-rogue-root-ca/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,6 +259,12 @@ Online tests
 * [AmIUnique](https://amiunique.org/) ([Source](https://github.com/DIVERSIFY-project/amiunique))
 
 
+Certificate Checks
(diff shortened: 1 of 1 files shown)


nodiscc:
> You are using the Firefox browser. Firefox has its own certificate store and is therefore not affected by this problem. You should repeat this test with the Internet Explorer, Edge or Chrome.

pyllyukko:
Yep. Even though this was (again) bad news regarding PKI, as it doesn't affect us Firefox users, I don't think we need to mention this.

PR #88, opened 2015-11-30
added Pref Helper

CHEF-KOCH:
Added an addon which might useful and not waste resources, it's not even 2 KB big ...


pyllyukko:
I don't see the added value of mentioning this in the README.

PR #90, opened 2015-12-15
Firefox Addon Detector added

CHEF-KOCH:
Added an online test which detects (or in our case not detect) addons. It's unofficial one and needs Javascript enabled + Session cookies. Remember that if you use NoScript and uBlock to 'unlock' (allow it/whitelist) to get some results. If it shoes something this means you have a leak which allows external pages to detect your addons, but we have 'normally' fixed that with the addon meta-data blocking thing.


pyllyukko:
It's already there on [L252](https://github.com/pyllyukko/user.js/blame/master/README.md#L252).

PR #95, opened 2016-01-04
Add a note around third-party cookies and payment gateways

fmarier:
Realistically, a lot of people will need to enable third-party cookies
to unbreak various legacy websites.

Therefore, let's make sure they at least expire at the end of the
browsing session if they aren't disabled altogether.

### DIFF
--- a/user.js
+++ b/user.js
@@ -352,7 +352,11 @@ user_pref("network.http.sendSecureXSiteReferrer",		false);
 
 // CIS 2.5.1 Accept Only 1st Party Cookies
 // http://kb.mozillazine.org/Network.cookie.cookieBehavior#1
+// This breaks a number of payment gateways so you may need to comment it out.
 user_pref("network.cookie.cookieBehavior",		1);
+// Make sure that third-party cookies (if enabled) never persist beyond the session.
+// https://feeding.cloud.geek.nz/posts/tweaking-cookies-for-privacy-in-firefox/
+user_pref("network.cookie.thirdparty.sessionOnly",      true);
 
 // user-agent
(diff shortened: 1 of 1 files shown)


nodiscc:
Hi @fmarier I have been browsing with third-party disabled for _years_ without a single issue. I also set this on computers I deploy and have received zero complaints from users. Third party cookies can also be very invasive. Can you share some examples of how/when/how often it breaks sites?

fmarier:
I ran into this problem twice just in the last 2 weeks: `justeat.ca` and `orleansexpress.com`

It's not possible to make a purchase on those sites without enabling third-party cookies. It doesn't even work if you set allow third-party cookies from "visited" sites.

pyllyukko:
Thanks. Maybe we should add a note about this also into the README's [known problems](https://github.com/pyllyukko/user.js#known-problems) section...

fmarier:
> Thanks. Maybe we should add a note about this also into the README's known problems section...

Good point, I'll add something and send a new PR.

pyllyukko:
> Good point, I'll add something and send a new PR.

Thanks!

I concur with @nodiscc, as I also haven't had any issues without 3rd party cookies, but it's good to have this covered nevertheless.

fmarier:
> Good point, I'll add something and send a new PR.

That's #96.

dveditz:
I also browser with 3rd party cookies disabled and have since the last millennium. Mostly it works fine for read-only web use, but for some people using some sites it will be a daily or weekly breakage. I vaguely remember needing to enable it to buy something from target.com a few years back. I had to whitelist login.persona.org in order to sign into any site using Mozilla's Persona single sign-on mechanism--not that popular, but if you use it you need it and the failures can be mysterious. Likewise I had to whitelist the SSO site used by my corporate network.

PR #96, opened 2016-01-04
Document payment gateway breakage

fmarier:
As requested on https://github.com/pyllyukko/user.js/pull/95#issuecomment-168805641


pyllyukko:
Thanks.

Issue #103, opened 2016-01-17
Split up user.js


szepeviktor:
Please consider splitting up `user.js` to sections into `user.js.d` directory and add a build script:
`cat user.js.d/* > user.js`

szepeviktor:
- 000-header.js               
- 001-html-apis-dom.js        
- 002-misc.js                 
- 003-extensions-plugins.js   
- 004-features-components.js  
- 005-automatic-connections.js
- 006-http.js                 
- 007-caching.js              
- 008-ui-related.js           
- 009-tls-http-ocsp.js        
- 010-ciphers.js

CHEF-KOCH:
Why separate if we get the same into one file?

A lot of stuff is obsolete/deprecated I highly suggest (as mentioned in other threads) to remove this first instead of 'split' it.

szepeviktor:
Actually I cannot see through one file but these section can be handles.
Now I examine all of the prefs and I decide keeping/changing them.

CHEF-KOCH:
I don't see any benefit because according to you script you execute all of these settings it would make sense if you e.g. want to exclude obsolete ciphers settings (which are not needed anymore in latest FF releases). But in such cases we comment out some possible 'problematically' lines.

szepeviktor:
Do you know Debian's `/etc/apt/sources.list.d/`? Its purpose is to keep the code clean.

CHEF-KOCH:
APT isn't avl. for MAC/Win and this project should work for all not only Linux so. You can do the same with wget which is avl. for all OS to get latest user.js and place it automatically into your profiledir.

szepeviktor:
Excuse me, I wasn't clear. The distributed files would stay the same, only the *development version* would be in `user.js.d`.

pyllyukko:
> Please consider splitting up user.js to sections into user.js.d directory and add a build script:
> cat user.js.d/* > user.js

Somehow I feel it's easier to keep them all in one. At least when it's still reasonable sized. When it exceeds some 1000+ lines or something like that, it might be better to split it, but at this point I think it's better all in one.

Also it would be one more extra step to add changes to it :)

CHEF-KOCH:
I only would suggest to remove as much as possible comments and better link into the threads (if present) or explain it into the wiki which would it make easier:

* Do Download/fork the documentation since wiki is forkable
* Reduce the size 
* Since no one or only few people reading documents/comments it would help to get an cleaner overview
[...]

PR #104, opened 2016-01-18
update user.js to handle #103

CHEF-KOCH:
- shrink and reduce the size by remove the comments and explanation since it's given in the Readme & Wiki + commit changeset


pyllyukko:
I don't want to be cross-referencing the README file all the time. I think the links are necessary.

PR #107, opened 2016-01-27
Update user.js

CHEF-KOCH:
* Ghostery removed it's not necessary with Mozilla's internal blocklist (from Disconnect)


nodiscc:
@pyllyukko what does `extensions.ghostery.privateBrowsing` do (can't find any documentation), and can it be moved to [extensions / plugins](https://github.com/pyllyukko/user.js/blob/master/user.js#L234) section?

Or can this pref be safely removed?

Or can we remove Ghostery from the add-ons list (other issue)?

PR #108, opened 2016-01-29
Update user.js - DuckDuckGo

CHEF-KOCH:
(optional) we added DuckDuck Go as default search engine but it still uses Google as first search so let's fix this
(optional2) let (in case you comment out) the url search should use DuckDuckGo by default

### DIFF
--- a/user.js
+++ b/user.js
@@ -102,7 +102,7 @@ user_pref("clipboard.autocopy",		false);
 // URL instead of asking from google.
 // http://kb.mozillazine.org/Keyword.enabled#Caveats
 user_pref("keyword.enabled",		false);
-//user_pref("keyword.URL", "https://duckduckgo.com/html/?q=!+"");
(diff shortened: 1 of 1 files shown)


pyllyukko:
> (optional) we added DuckDuck Go as default search engine but it still uses Google as first search so let's fix this

How does this work? I also have ```browser.search.order.1 = Google```, but Firefox still uses DDG.

CHEF-KOCH:
`keyword.URL` also needs to be `https://duckduckgo.com/html/?q=!+` instead of `http://www.google.com/search?q=`. 


Also ensure/reset the following things:
* browser.search.selectedEngine (removed)
* browser.search.defaultenginename needs (if not set to `DuckDuckGo (SSL)`)
* browser.safebrowsing.provider.0.name (removed)
* and this is the important one -> `browser.search.order.US.1`  which I think is set for you as `data:text/plain,browser.search.order.US.1=Google` this maybe needs set for you to DuckDuckGo but for me default it works.

PR #109, opened 2016-01-29
Social stuff

CHEF-KOCH:
* Kill all social stuff (yes we are not social) :stuck_out_tongue_winking_eye: 
* Do not allow remote installations of addons
* Kill the share it button
* We not want any whitelist of social stuff

### DIFF
--- a/user.js
+++ b/user.js
@@ -271,6 +271,14 @@ user_pref("browser.safebrowsing.downloads.remote.enabled",	false);
 // https://support.mozilla.org/en-US/kb/save-web-pages-later-pocket-firefox
 user_pref("browser.pocket.enabled",		false);
 
+// Kill all social related addons
+user_pref("social.whitelist", "");
+user_pref("social.manifest.facebook", false);
+// Do not allow remote installation in the background without permissions
+user_pref("social.remote-install.enabled", false);
+// Kills the share it button
+user_pref("social.share.activationPanelEnabled", false)
+
 /******************************************************************************
  * automatic connections                                                      *
  *                                                                            *


nodiscc:
Hello, I am mostly fine with this change, but was wondering if there is a real benefit in changing this. AFAIK this feature does not leak any information without explicit user action, and provides a clean, browser-based way to share stuff with web services. Many open source projects make use of this feature (eg. https://github.com/shaarli/Shaarli/ has Firefox Share integration).

CHEF-KOCH:
Both of the social api and the share activity isn't isolated or sandboxed and uses 3rth party libaroes for 3rd üarties like chat and video (which we disabled), so practically an infected addon could abuse this to read out metadata and this is why I bumped it. In the past Googles/Facebook's buttons set an own cookie which monitored your surfing behaviors as an example. 

I don't not enough about the share button itself but I think it's better to block it by default to reduce metadata as much as it possible. 

My research are based on:
* https://blog.mozilla.org/futurereleases/2012/10/22/help-us-test-the-social-api-with-facebook-messenger-for-firefox/
* https://blog.mozilla.org/futurereleases/2013/06/27/social-api-for-all/
* https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Social_API

Hunter-Github:
@nodiscc - think there should be some disclosure that you're working on shaarli.

nodiscc:
Ok, yes I used to work on Shaarli (much less nowadays), for context the share button proved itself a good alternative to bookmarklets (they do not work everywhere because of CSP, and allowing javascript bookmarks could also be a security risk). For people who _do_ really want to share items to web services, I think the Firefox Share button is also a decent alternative to allowing embedded "share to.." buttons in web pages (which silently leak browsing data to these services).

But I have no strong opinion on this. Mozilla's list of providers is also quite restricted and I don't think this is used much (the usability tradeoff is small).

CHEF-KOCH:
I don't see it, why anyone needs a share button on desktop, on mobile I would maybe agree but not on desktop

nodiscc:
May I ask why?

CHEF-KOCH:
Well you're not objective enough because you work on this and then of course you use and _need_ it, in fact strg + c and strg + v is much quicker then any tool that possible comes with negative side-effects but again I'm not really involved in this social and sharing thing but as long no one can show me that this isn't involved in the social api (which is known to be _problematic_) I not change my pull request. The goal is to reduce the traffic as much as possible and harden FF, to use social media and such external addons/share thing is exactly the opposite because then we can use FF out of the box, install uMatrix and block everything and get the same .... and this is what the creator not want (I think).

nodiscc:
> in fact strg + c and strg + v is much quicker

For me the 2 competing workflows are, using share button (or bookmarklet):

 * Select text to use as description
 * click Share button
 * click `Save`

Or using Ctrl+C Ctrl+V:

 * Open my link sharing tool in another tab
 * Switch back to the tab containing the page I want to share
 * Select URL in the address bar
 * `Ctrl+C`
 * Switch back to link sharing tool tab
 * Click `Add Link`
 * `Ctrl+V` URL in the URL field, hit Enter
 * Switch back to the tab containing the page I want to share
 * Select text to use as description
 * `Ctrl+C`
 * Switch back to the tab containing the page I want to share
 * Select description field
 * `Ctrl+V`
 * Click `Save`

The second one is impractical. As I said it might be a feature that is actually used by few people (maybe Mozilla has metrics on this huehuehue), so no big harm in disabling it (it can be re-enabled) but if you can provide some actual examples of how this feature is _problematic_, it would be nice.

pyllyukko:
Isn't the social & share functionality completely opt-in?

CHEF-KOCH:
You're right according to Mozilla this social thingy is (all) opt-in from now.

pyllyukko:
Ok, so since it's an opt-in feature and users might actually want/need/use this, I'll close this.

pyllyukko:
FYI. I noticed that I'm unable to use the whole Firefox Share at all (the button is grey) as long as I have ```browser.privatebrowsing.autostart``` set.

PR #111, opened 2016-02-01
Disable WebPush?

CHEF-KOCH:
Disable WebPush which is useless at the moment and in the next coming releases because this is possible dangerous

### DIFF
--- a/user.js
+++ b/user.js
@@ -7,6 +7,15 @@
  * HTML5 / APIs / DOM                                                         *
  *                                                                            *
  ******************************************************************************/
+ 
+// Disable Web push:
+// https://en.wikipedia.org/wiki/Push_technology
+// https://www.ietf.org/mail-archive/web/webpush/current/threads.html#00182
+user_pref("dom.push.connection.enabled", false);
+user_pref("dom.push.enabled", false);
+user_pref("dom.push.serverURL", "");
+user_pref("dom.push.udp.wakeupEnabled", false);
+user_pref("dom.push.userAgentID", "");
 
 // disable Location-Aware Browsing
 // http://www.mozilla.org/en-US/firefox/geolocation/


pyllyukko:
Might be dangerous, yes. But this might be one of those technologies we need to allow (like DOM localStorage), once the web app developers pick up on this.

CHEF-KOCH:
Is it? I'm asking because to clarify the question about the social networks, because from what I know only whatsapp, hello and such using this. If we support such things then I will ignore in the future such changes.

pyllyukko:
If major social media players are using this, then it's more the reason not to block them (if it breaks functionality that is).

We want to stay compatible with sites.

But I don't have enough knowledge about this technology and it's prevalence to decide at this point. I think we'll let this open and up to discussion ATM.

Gitoffthelawn:
I agree with @CHEF-KOCH that it's best not to have Web Push right now.  I also agree with @pyllyukko that this is likely to be one of those technologies that will be important once it is more widely used.

I dislike the idea of Web Push, but I think it will generate profit for some large companies, thus it will probably become popular.

CHEF-KOCH:
I will do now some research and ask Mozilla what is the plan of this and if there are reasons to enable this by default instead of add an whitelist (similar to social.whitelist) to proper handle it.

Gitoffthelawn:
@CHEF-KOCH +1 for a whitelist

pyllyukko:
Related Tor Browser issue: <https://trac.torproject.org/projects/tor/ticket/18801>

nodiscc:
> Might be dangerous, yes. But this might be one of those technologies we need to allow (like DOM localStorage), once the web app developers pick up on this.

Same discussion we had in https://github.com/pyllyukko/user.js/commit/c65632f066223516d26ea67aecfa7a9fb04a3238

What I suggest:

```diff
+ 
+// Disable Web Push
+// https://en.wikipedia.org/wiki/Push_technology
+// https://trac.torproject.org/projects/tor/ticket/18801
+// https://support.mozilla.org/t5/Basic-Browsing/Web-Push-notifications-in-Firefox/ta-p/28744
+// https://support.mozilla.org/t5/Firefox/How-to-stop-the-webpush-notifications/m-p/1292770
+// https://developer.mozilla.org/en/docs/Web/API/Push_API
+// https://github.com/chrisdavidmills/push-api-demo
+user_pref("dom.push.connection.enabled", false);
+user_pref("dom.push.enabled", false);
+user_pref("dom.push.serverURL", "");
+user_pref("dom.push.udp.wakeupEnabled", false);
+user_pref("dom.push.userAgentID", "");
```

pyllyukko:
> What I suggest:

> +// Disable Web Push
> +// https://en.wikipedia.org/wiki/Push_technology
> +// https://trac.torproject.org/projects/tor/ticket/18801
> +// https://support.mozilla.org/t5/Basic-Browsing/Web-Push-notifications-in-Firefox/ta-p/28744
> +// https://support.mozilla.org/t5/Firefox/How-to-stop-the-webpush-notifications/m-p/1292770
> +// https://developer.mozilla.org/en/docs/Web/API/Push_API
> +// https://github.com/chrisdavidmills/push-api-demo
> +// Defaults: true, true, wss://push.services.mozilla.com/, ""
> +user_pref("dom.push.connection.enabled", false);
> +user_pref("dom.push.enabled", false);
> +user_pref("dom.push.serverURL", "");
> +user_pref("dom.push.udp.wakeupEnabled", false);
> +user_pref("dom.push.userAgentID", "");

We should use the master switches only and leave rest of the settings alone. Maybe it requires some additional testing, but disabling the service workers might be enough.

Also I found few more links on the topic:
* [Web Push Arrives in Firefox 44](https://hacks.mozilla.org/2016/01/web-push-arrives-in-firefox-44/) (see also the "What about privacy?" part)
* [Simple Push](https://developer.mozilla.org/en-US/docs/Archive/Firefox_OS/API/Simple_Push_API)... this actually explained a lot to me as I have used push notifications without service workers and wondered what's that all about
* (Yet another) [How to disable web push notifications in Firefox?](https://support.mozilla.org/t5/Firefox/How-to-disable-web-push-notifications-in-Firefox/m-p/1281001)

nodiscc:
This patch is about actively disabling webpush and preventing connections to the central Mozilla Push server, not just disabling/masking notifications.

Tor browser has merged these patches: https://trac.torproject.org/projects/tor/attachment/ticket/18801/0001-Bug-18801-disable-the-dom.push-prefs.patch, https://trac.torproject.org/projects/tor/attachment/ticket/18801/0001-Bug-18801-clear-the-DOM-Push-server-URL.patch so it would make sense to do the same.

`dom.webnotifications.enabled` and `dom.webnotifications.serviceworker.enabled` can be added to the UI section, but it would be good to know which of these prefs is the "master switch" for push connections.

@fmarier maybe you have more info on this?

PR #112, opened 2016-02-01
Remove default website protocol handlers

CHEF-KOCH:
No one (and I'm sure ever) use this, gmail is comment out because it could be useful if you use webmail but the other services are not widley used or there are better extension for it

### DIFF
--- a/user.js
+++ b/user.js
@@ -452,6 +452,23 @@ user_pref("browser.helperApps.deleteTempFileOnExit",		true);
 // https://developer.mozilla.org/en-US/docs/Mozilla/Preferences/Preference_reference/browser.pagethumbnails.capturing_disabled
 user_pref("browser.pagethumbnails.capturing_disabled",		true);
 
+// Remove default website protocol handlers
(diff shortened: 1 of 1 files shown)


pyllyukko:
Do these settings matter at all?

Gitoffthelawn:
I think some people use Mibbit for IRC.

CHEF-KOCH:
On external addons you can simply implement your own protocol handler to add e.g. irc, so I see no problem there is also an Mozilla page which explains how to add protocol handlers to your extension/addon.

nodiscc:
I have other prefs in place for this:

```js
user.js:user_pref("network.protocol-handler.app.irc", "/usr/bin/transmission-gtk");
user.js:user_pref("network.protocol-handler.app.magnet", "/usr/bin/transmission-gtk");
user.js:user_pref("network.protocol-handler.expose.sftp", false);
user.js:user_pref("network.protocol-handler.external.irc", true);
user.js:user_pref("network.protocol-handler.external.magnet", true);

```

Whereas you use `gecko.handlerService.schemes.*`. Which one is correct?

Also I think this falls into the "Personal user preferences" category and could be discarded.

PR #113, opened 2016-02-01
Finally killed newtab

CHEF-KOCH:
* Corrected directory.source 
* added annyoing intro skip


pyllyukko:
I prefer the ```data:text/plain,{}```.

PR #114, opened 2016-02-01
Kill SHA1 certificates

CHEF-KOCH:
Firefox now allows SHA1 certificates issued after Jan 1 2016 again because HTTPS scanning/MitM products [are outdated](https://blog.mozilla.org/security/2016/01/06/man-in-the-middle-interfering-with-increased-security/).

```
You can influence Firefox's policy on this by setting "security.pki.sha1_enforcement_level":
0 = allow SHA-1
1 = forbid SHA-1
2 = allow SHA-1 only if not before < 2016-01-01
```

### DIFF
--- a/user.js
+++ b/user.js
@@ -555,6 +555,7 @@ user_pref("security.enable_ssl3",		false);
 // https://wiki.mozilla.org/SecurityEngineering/Public_Key_Pinning#How_to_use_pinning
 // "2. Strict. Pinning is always enforced."
 user_pref("security.cert_pinning.enforcement_level",		2);
+//user_pref("security.pki.sha1_enforcement_level", 1);
 
 // https://wiki.mozilla.org/Security:Renegotiation#security.ssl.treat_unsafe_negotiation_as_broken
 // see also CVE-2009-3555
--- a/user.js
+++ b/user.js
@@ -555,6 +555,9 @@ user_pref("security.enable_ssl3",		false);
 // https://wiki.mozilla.org/SecurityEngineering/Public_Key_Pinning#How_to_use_pinning
 // "2. Strict. Pinning is always enforced."
 user_pref("security.cert_pinning.enforcement_level",		2);
+
+// Kill SHA1 certificates
+// https://bugzilla.mozilla.org/show_bug.cgi?id=942515#c32
 //user_pref("security.pki.sha1_enforcement_level", 1);
 
 // https://wiki.mozilla.org/Security:Renegotiation#security.ssl.treat_unsafe_negotiation_as_broken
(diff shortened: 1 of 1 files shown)


fmarier:
Given that the main issue which forced the revert of this patch was the presence of enterprise MITM software (or anti-virus), perhaps it should be enabled by default in user.js?

CHEF-KOCH:
I use this now 3 weeks without any troubles, bigger pages are frequently updating there certs and I not found any other forum/page which gave me troubles.

fmarier:
My suggestion would be:

- default to `enforcement_level = 2` since it should work fine for all those who use `user.js`
- have `enforcement_level = 1` commented-out for those who want to disable SHA-1 completely even though it's still allowed as long as the cert was issued prior to 2016.

I'm currently connecting to github via a wifi gateway with a valid (issued in 2014) SHA-1 cert and so `enforcement_level = 1` is likely to cause lots of problems if it's the default.

CHEF-KOCH:
Just increase timeout and you not get any problems on github.

fmarier:
I'm not talking about github itself, I'm talking about the login page for the wifi I'm using. That's using a valid SHA-1 cert.

CHEF-KOCH:
Ah okay sorry, seems I misunderstand you. :disappointed:

pyllyukko:
> default to enforcement_level = 2 since it should work fine for all those who use user.js

This sounds like a good default. I'll start testing with ```1``` and see whether it causes problems.

pyllyukko:
When ```security.pki.sha1_enforcement_level``` is set to ```1```, you'll get an ```sec_error_cert_signature_algorithm_disabled``` error. But it's still possible to bypass it and add an exception, so it might be safe to set this to ```1``` by default.

md-2

fmarier:
https://groups.google.com/a/chromium.org/d/msg/security-dev/QNVVo4_dyQE/U_nMQWDnGqYJ is the latest update from Chrome, Mozilla will likely follow a similar deprecation/removal plan.

Until SHA-1 is officially sunset, it might be a bit annoying to have to rely on exceptions to whitelist valid certs. When you're dealing with embedded third-party resources, you don't always see the cert error page which allows you to add exceptions. I suspect some (crappy) payment gateways might rely on these kinds of certs for compatibility with very old browsers/OSes.

So personally, I would make level 2 the default for now and reassess later during the year as things progress.

CHEF-KOCH:
It's not annoying because most pages already dropping SHA1 and it's now known for several years that this is weak, how long we always should wait? It's a possible security issue we talking about, there is (for me) no compromise in that, as I already said in other threads just ask your webmaster to fix that (takes 5 minutes).

And remember bigger pages like Google, Facebook, GitHub, Schneier, Instagram often change the certificates or monitor on such problems so I still see no problem, we only talking then about pages with only a few members or as mentioned CapitalPortal vulnerable login pages but if you see such a warning this is exactly what we want because there is no other indicator to mark stuff as possible problematic, for me this is a perfect situation right now.

pyllyukko:
Just an update:

I ran my Firefoxes with this setting set to ```1``` and bumped into warnings way too often. So yes, it is best we have this at ```2``` for now.

CHEF-KOCH:
Official Update on this:
* https://blog.mozilla.org/security/2016/02/24/payment-processors-still-using-weak-crypto/
* https://www.schneier.com/blog/archives/2015/10/sha-1_freestart.html
* https://groups.google.com/forum/#!msg/mozilla.dev.security.policy/RHBHXJOG8Io/FJuaWeXAAQAJ
* https://groups.google.com/forum/#!topic/mozilla.dev.security.policy/RHBHXJOG8Io

ALL SHA-1 certs killed except:
* Symantec due the fact they renewed the Cert 31. Dec. 2015 + most if not all creditcard reader are not sha1 ready


R.I.P. Symantec + SHA1 :+1: :skull:

PR #116, opened 2016-02-05
postprocessing.sh created to use Firefox GCLI

Hunter-Github:
GCLI is better than an external user.js file if you want to keep all your settings in one place (`about:config`). The script works on Linux, not tested on OS X or Windows. Firefox GCLI allows one to set preferences from the command line inside the browser.

### DIFF
--- a/postprocessing.sh
+++ b/postprocessing.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# The script converts `user.js` into a sequence of commands for Firefox development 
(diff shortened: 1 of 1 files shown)


CHEF-KOCH:
I don't think we need it, in the comments oft this script it tells about 'in development console' but we want user.js before you load firefox the first time (in case of a fresh installation). 

But I could merge it then in my addon to parse/paste directly into the console if there is a wish for such behavior.

Hunter-Github:
Okay, fine by me.

PR #118, opened 2016-02-15
Deactivates the graphical tour on Australis UI

CHEF-KOCH:
This deactivates the graphical Tour you see the first time (only with Australis UI enabled).


nodiscc:
Can be closed, whe have `user_pref("browser.uitour.enabled",             false);`

PR #119, opened 2016-02-16
Remove redundant/obsolete ToDo entries

CHEF-KOCH:
- HTML 5 canvas can only be disabled or enabled
- Fingerprinting can be avoided by using Canvas Blocker or NoScript which then requires to allow the page (you will see popup, same like NoScrits one click protection)
- Tor (Browser) is fixed with https://gitweb.torproject.org/tor-browser.git/patch/?id=4da1f32
- Pixel Perfect is the same as BrowserLeaks, it uses pixel canvas to get e.g. OS Information which will be pares by the web page

Timezone Spoofing
- Requires script injection aka Random Agent spoofer

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,15 +391,10 @@ TODO
 ----
 
 - [ ] [HTML5 canvas](https://en.wikipedia.org/wiki/Canvas_element)
-  - [Canvas fingerprinting](https://en.wikipedia.org/wiki/Canvas_fingerprinting)
-  - [BrowserLeaks.com](https://www.browserleaks.com/canvas)
-  - [Mozilla official Canvas explanation](https://developer.mozilla.org/en-US/docs/HTML/Canvas)
-  - [Tor fingerprinting topic](https://www.torproject.org/projects/torbrowser/design/#fingerprinting-linkability)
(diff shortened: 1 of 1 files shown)


nodiscc:
This is part of #220 but let's move the reference links to the canvas section  in `user.js` instead of just deleting them.

Let's also move Random Agent Spoofer/Timezone Spoofing research to a new issue

@pyllyukko would you be ok with this?

pyllyukko:
> This is part of #220 but let's move the reference links to the canvas section in user.js instead of just deleting them.
> 
> Let's also move Random Agent Spoofer/Timezone Spoofing research to a new issue

IMO the timezone is already resolved in 859b7b62407c8df73b4d718462a60bd68a0d36ce.

> @pyllyukko would you be ok with this?

Sounds good.

pyllyukko:
Or should we just remove the whole TODO section and not clutter the user.js? Not sure if all the links are even relevant/up-to-date anymore. The information is available in the git log :smile:

nodiscc:
See https://github.com/pyllyukko/user.js/pull/224

nodiscc:
Can be closed

PR #121, opened 2016-02-21
Disable Hardware Acceleration to avoid fingerprints

CHEF-KOCH:
* Not needed since FFmpeg 3.0 is here, which allows hardware acc. anyway even if you disable it directly with about:config (just use right click to enable it if you want it) of course the page must support this but I'm sure YouTube/Vimeo/MetaCafe and others will change this asap ...
* This needs to be disabled also (if in use) in Adobe Flash/Macromedia to avoid Canvas fingerprinting for font's
* Hardware Acc. doesn't perform well on some systems, which should better controlled by driver settings to override it (if you want)

Research:
* http://www.phoronix.com/scan.php?page=news_item&px=FFmpeg-3.0-Released
* https://hacks.mozilla.org/2010/09/hardware-acceleration/
* https://www.google.com/patents/US20120215896 (just an example for device identification on networks)
* .... several already mentioned Tor papers about this

### DIFF
--- a/user.js
+++ b/user.js
@@ -2,6 +2,17 @@
  * user.js                                                                    *
  * https://github.com/pyllyukko/user.js                                       *
  ******************************************************************************/
+ 
+ /******************************************************************************
+ * Avoid hardware based fingerprintings                                       *
+ * Canvas/Font's/Plugins                                                                           *
+ ******************************************************************************/
+// https://wiki.mozilla.org/Platform/GFX/HardwareAcceleration
+// https://www.macromedia.com/support/documentation/en/flashplayer/help/help01.html
+//https://github.com/dillbyrne/random-agent-spoofer/issues/74
+ user_pref("gfx.direct2d.disabled",		true);
+ user_pref("layers.acceleration.disabled",		true);
+ 
 
 /******************************************************************************
  * HTML5 / APIs / DOM                                                         *


pyllyukko:
How does this help? I don't notice any changes in the [canvas fp test](https://www.browserleaks.com/canvas), when enabling these settings. Can you elaborate a bit?

CHEF-KOCH:
Because your mentioned test doesn't test hardware accelerated things. [I only found an PDF paper](https://network23.org/inputisevil/2015/09/06/how-html5-apis-can-fingerprint-users/) which is very complex which explains the hardware acc. fingerprinting, especially on mobile devices and Browsers on HTML5 players because this directly works in memory and then the attacker possible gets the ability to 'fingerprint' /control/attack/whatever the microphone/sensors and other things. 

On [Wikipedia](https://en.wikipedia.org/wiki/Device_fingerprint) this is a bit better explained because it explains under active fingerprinting that DRM and others are affected by this, ressource timing and 'the clock' problem can be used to identify e.g. driver version (but these was fixed from what I know).

On [TorTracker](http://tor-talk.torproject.narkive.com/453ioKWX/hardware-acceleration-ok-or-not) I think they mentioned that this avoids time based attacks. I also had a paper about this but I can't find it (yet) but if I find this then I add it into the pull directly.

pyllyukko:
Yeah, it's probably good to have these set. Thanks!

PR #123, opened 2016-02-21
add Profile support

CHEF-KOCH:
Profiles for different reasons added

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,18 @@ It is common for [client side attacks](https://www.offensive-security.com/metasp
 * Enables Firefox's built-in [tracking protection][12]
 * Disables [telemetry](https://wiki.mozilla.org/Telemetry), [Crash Reporter](https://support.mozilla.org/en-US/kb/Mozilla%20Crash%20Reporter), [healt report](https://support.mozilla.org/en-US/kb/firefox-health-report-understand-your-browser-perf), [heartbeat](https://wiki.mozilla.org/Advocacy/heartbeat) and other such privacy invading nonsense
 
+### Profiles
+
+You can start Firefox with [different profiles](http://kb.mozillazine.org/Profile_folder_-_Firefox) for [different reasons](https://support.mozilla.org/en-US/kb/profiles-where-firefox-stores-user-data?redirectlocale=en-US&redirectslug=Profiles), e.g. to test a page without extensions or for use  it for online-banking. An benefit is that you can Sandbox these profiles with LXC/Akrose and/or combine it with AppAmor/SELinux to start this in it's own instance. 
+
+| Profile | Description | 
+| :--- | :----: | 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -100,6 +100,18 @@ It is common for [client side attacks](https://www.offensive-security.com/metasp
...
+### Profiles
+
+You can start Firefox with [different profiles](http://kb.mozillazine.org/Profile_folder_-_Firefox) for [different reasons](https://support.mozilla.org/en-US/kb/profiles-where-firefox-stores-user-data?redirectlocale=en-US&redirectslug=Profiles), e.g. to test a page without extensions or for use  it for online-banking. An benefit is that you can Sandbox these profiles with LXC/Akrose and/or combine it with AppAmor/SELinux to start this in it's own instance. 
+
+| Profile | Description | 
+| :--- | :----: | 
+| firefox -P <profile-name> -no-remote  | Start the profile in it's own instance  | 
+| firefox -P <profile-name> -no-remote -safe-mode  | + deactivates themes and extensions | 
+| firefox -P <profile-name> -no-remote -private  | + private mode (alias incognito mode) |
+| firefox -P <profile-name> -no-remote -safe-mode -private | + secure/private mode |
```
fmarier: `safe-mode` in Firefox is analogous to the Windows safe mode: if you're having problems with things crashing and you're unable to start Firefox, you can start it in safe-mode and hopefully solve the problem there with most things disabled.

I agree with the first description ("deactivates themes and extensions") but I'm not sure about calling it "secure". Safe Browsing is [disabled in safe mode](https://bugzilla.mozilla.org/show_bug.cgi?id=1134954) for example, which means that users are vulnerable to malware, phishing and that tracking protection is not active.

I would personally not encourage people to use safe-mode because it's really meant to be a recovery feature. If you don't want extensions and plugins in a profile, you can just disable them in about:addons.

pyllyukko: I agree. Also, it is missing the ```--ProfileManager``` switch. I think running Firefox with ```--ProfileManager --no-remote``` is the only thing that needs to be documented in this context, as the user.js takes care of keeping the browser in private browsing mode.

The "LXC/Akrose" and "AppAmor/SELinux" parts should have links to these technologies.

nodiscc: I would remove the table, and link to http://kb.mozillazine.org/Command_line_arguments

nodiscc: I would also rewrite the paragraph as 

> Firefox supports running different [profiles](https://support.mozilla.org/t5/Install-and-Update/Profiles-Where-Firefox-stores-your-bookmarks-passwords-and-other/ta-p/4608), each with its own preferences, add-ons, bookmarks, history, and more. The [Profile directory](http://kb.mozillazine.org/Profile_folder_-_Firefox) contains all information about a profile. To enable the Profile Manager, run Firefox with command-line arguments `--no-remote --P`

> See http://kb.mozillazine.org/Command_line_arguments

PR #130, opened 2016-03-13
add anticanvas

CHEF-KOCH:
Seems the better deal?

.... compared to the 'old' canvas addon ... (need confirmation)


README.md
```diff
@@ -398,6 +398,7 @@ TODO
   - [Bug 967895](https://bugzilla.mozilla.org/show_bug.cgi?id=967895)
   - [Pixel Perfect: Fingerprinting Canvas in HTML5](http://www.w2spconf.com/2012/papers/w2sp12-final4.pdf)
   - https://addons.mozilla.org/en-US/firefox/addon/canvasblocker/, https://github.com/kkapsner/CanvasBlocker/issues/
+  - https://addons.mozilla.org/de/firefox/addon/anticanvasfingerprinting/
```
fmarier: You should remove the `/de` from the URL to avoid sending people to the German version of AMO.

In fact, we should also remove the `/en-US` from all of the other links so that people get AMO in their language automatically.

Issue #38, opened 2015-08-10
[Talk] Canvas fingerprinting


CHEF-KOCH:
This thread is designed to talk about and find possible workarounds to handle the [canvas problem](https://en.wikipedia.org/wiki/Canvas_fingerprinting) (see also Readme.md  todo).

The sad news are that there is nothing much we can do without any bigger modifications. Canvas are more or less hard-coded into the Browser/HTML5/Headers to provide certain detection ability's. 

Research:
- [Font enumeration](http://www.lalit.org/lab/javascript-css-font-detect/)
- [akademie.dw.de](http://akademie.dw.de/digitalsafety/your-browsers-fingerprints-and-how-to-reduce-them/)
- [About canvas fingerprinting](http://www.propublica.org/article/meet-the-online-tracking-device-that-is-virtually-impossible-to-block)
- [Wikipedia article about canvas fingerprinting](https://en.wikipedia.org/wiki/Canvas_fingerprinting)
- [Tor patches](https://gitweb.torproject.org/torbrowser.git/tree/src/current-patches/firefox)
- [Tracking using ETags](https://en.wikipedia.org/wiki/HTTP_ETag#Tracking_using_ETags)
- [Evercookies](https://github.com/samyk/evercookie#browser-storage-mechanisms)
- [Academic research paper describing canvas fingerprinting (.pdf)](https://securehomes.esat.kuleuven.be/~gacar/persistent/the_web_never_forgets.pdf)
- [Partial database of websites that have used canvas fingerprinting](https://securehomes.esat.kuleuven.be/~gacar/sticky/index.html)
- [Browser test page](https://www.browserleaks.com/canvas)
- [PanOptiClick test page](https://panopticlick.eff.org/)
- [X-Forwarded-For](http://en.wikipedia.org/wiki/X-Forwarded-For)
- [HTTP header fields overview](http://en.wikipedia.org/wiki/List_of_HTTP_header_fields)
- [Fingerprinting can be performed against the NoScript whitelist](http://cseweb.ucsd.edu/~hovav/papers/mbys11.html)
- ISP/NSA use deep packages analyzers to identification the traffic (can also be done in almost real-time) [see WikiLeaks documents], encryption is the only solution but coasts more power and bandwidth, of course it's not always possible due lack of standard encryption protocols on the www and apps.  
[...]

PR #136, opened 2016-04-06
Add comment for blank pages on Android

licaon-kter:


pyllyukko:
What does this mean exactly? I don't get it.

licaon-kter:
Which part, my comment or what the issue is with that config as `true`?

pyllyukko:
What the issue is.

Roman-Nopantski:
maybe he's talking about `identity.mobilepromo.android` and by extension `identity.mobilepromo.ios`

CHEF-KOCH:
Not every Android supports hardware acceleration (due older hardware) and i was said in other commits that it might can be used to track users. No need to confuse people to add <code>On Android turning this on will make all the pages blank.</code> to it, especially because it's wrong.

licaon-kter:
I agree, ok, maybe rephrase it as `might` then?

Regarding canvas, won't it be better to recommend https://github.com/kkapsner/CanvasBlocker instead?

CHEF-KOCH:
We already recommend Canvasblocker + AntiCanvas addon (see readme). To totally block canvas blocks a lot of stuff and faking them with wrong infos or block per-page isn't yet possible without addons. 

But this is another topic which is already opened. 

It's also not 'might' ... just saying, this doesn't have much to do how pages are rendered. It's just that the hardware may perform some functions faster compared to the CPU.

pyllyukko:
> We already recommend Canvasblocker + AntiCanvas addon (see readme).

Well I wouldn't say we recommend those. They are just on the TODO list. If we're gonna talk about canvas stuff, I'd suggest we'd move to #38.

CHEF-KOCH:
Please close.

pyllyukko:
> Please close.

I still want to know what's this about. How are pages blank? Is it unusable? @licaon-kter: Can you provide some screenshots or some more detail?

licaon-kter:
Oh, I missed the recommendation, indeed you do.

Now, regarding this PR, I forgot to add the screenshots:
* `layers.acceleration.disabled TRUE` -> https://i.imgur.com/ubuImTh.jpg
* `layers.acceleration.disabled TRUE` (thumbnails view) -> https://i.imgur.com/COfo6PM.jpg
* `layers.acceleration.disabled FALSE` -> https://i.imgur.com/PSxIjQ6.jpg

CHEF-KOCH:
Which Android OS?
Which FF version, beta or stable?
Which phone?
Which platform?

licaon-kter:
Sony Xperia 5.1.1
Firefox 45.0.1 (stable) or Orfox 38

What do you mean by platform?

On Android, at least on my device, you can set it up only via `about:config`, although once you've disabled it (as you do now), you can't see any pages, not even internal ones, hence you need to either edit `prefs.js` (if you have root) or clear all app data.

CHEF-KOCH:
Which Xperia, there exist a lot of models (Z, Z5, 8, X10,...). I guess it's on ARM platform. 

Well, I tested it on Cm 13 with latest Android Beta and it works well if I set <code>layers.acceleration.disabled </code> to true.

licaon-kter:
Z3. Then it might be a ROM thing, so this PR might help someone (besides me).

pyllyukko:
Ok. Now I'm starting to understand :)

Maybe it would still be better to have this on the [known problems](https://github.com/pyllyukko/user.js#known-problems) part of the README instead?

CHEF-KOCH:
I don't know how often I now explained that is not any about:config related issue. :_onion_head_objection:

licaon-kter:
Yeah maybe, afaik, no one reads anything :), let alone the source, README might be easier for the search engines to grok and help someone somewhere.

pyllyukko:
I added a note to the README. @licaon-kter: Do you think this is sufficient?

And thanks for reporting this problem.

licaon-kter:
Yes, I was just writing kinda exactly the same text. Thanks :)

CHEF-KOCH:
No need to add this to anything as long there is no FF bugtracker or AOSP Issue ticket. Otherwise it confuses, as mentioned the information that this directly affects rendering is also wrong, since software is then used which is simply slower. Could also be connection issue, or or or ... 

GitHub uses HTML5 canvas, which have hard. acc. support, on Desktop FF you see the Canvas popup. on mobile this isn't visible because canvas limitation + the fact that they're can use software fallback. 


Please remove and close it.

licaon-kter:
It's not limited to GitHub, it was just a sample c'mon, any page (both web and internal like Addons) will be blank, not slow, not with aliased fonts, not with missing UTF-8 coding, just plain blank, and you can't access `about:config` to change anything back... since it's blank.

I've searched before posting this and I did not find any reference regarding Android, just for desktop, and the problem was the other way around, acceleration ON was troublesome.

I don't see why some piece of info could be harmful here in any way, it would have helped me so I did not end up deleting 3 profiles and going all _divide et impera_ on both `user.js` and `prefs.js` settings.

If this information is wrong, please direct me on what to test next, I'm interested in the *true* culprit.

CHEF-KOCH:
It's wrong that's all.

licaon-kter:
I can't just install AOSP/CM13/etc, I can't unlock the bootloader, so it will be rejected as ROM specific, which is perfectly fine, plenty of Android _versions_ go around being slightly broken in different ways.

Regarding a Firefox specific bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1146783

CHEF-KOCH:
We can't add for each 'problem' someone may have workaround's it's waste of time because it's may related to Browser or the OS itself. The other thing is that it confuses and may scare people away because they see the 'mass' on problems which mostly aren't really problems. 

You could just try to use chromium it uses similar rendering technique on android, if it works then it's 'your own problem'. In the time we talked now it's also already fixed or redundant if you just could try the Beta FF version. 

I highly suggest to open an ticket yourself and report instead of link to something which is may another story. It not even mention something about hardware acceleration.

Again remove from Readme. Makes no sense. It worked for me well, even on Stock on my S6.

nodiscc:
> related to Browser or the OS itself

Yes, so let's keep it in the doc (_I have $random problem -> Your android ROM is broken and we can't fix it_)

> may scare people away because they see the 'mass' on problems

It's technical doc and I think it should reference all annoying problems (someone bothered to open an issue here) - even better, it has a one-line workaround.

licaon-kter:
> We can't add for each 'problem' someone may have workaround's

Yes you can, that's the whole point of FAQs and _Known issues_ sections since like forever.

> You could just try to use chromium it uses similar rendering technique on android

And disable acceleration there too or what? You did understand that it works fine OOTB, unless I purposely/manually disable acceleration, right?

>  if you just could try the Beta FF version

I'll try it and come back here to signal a change or not.

> I highly suggest to open an ticket yourself and report instead of link to something which is may another story.

Yes, it sounds the same, it might not be. I'll see to that too, yet another account meh, getting used to GitHub and its hosted projects, one account for them all :P

> It worked for me

You do realize that does not mean anything. Hopefully you don't have any issue on your hardware, it would be pretty sad to hear that line from anyone else.

takiainen:
I have this same issue, and definetly think it should be mentioned in readme. I've been trying to figure out during last few days which one of these hundred or so settings breaks my browser, until I found this. 

I've got Oneplus One running darkobas MM ROM. I've tried 38 based Icecat mobile browser and latest Firefox from fdroid - both have same issue. 

Everything is just blank, can't access about:config either.

PR #138, opened 2016-04-09
Jit was re-written for security reasons

CHEF-KOCH:
See [here](http://jandemooij.nl/blog/2015/12/29/wx-jit-code-enabled-in-firefox/). ESR/Android versions get the same changes with the next final build.


pyllyukko:
Let's wait and see how this evolves.

nodiscc:
@pyllyukko JIT bugs apparently fixed, see https://bugzilla.mozilla.org/show_bug.cgi?id=1215479

I think it is ok to remove this pref.

Issue #145, opened 2016-05-20
Another fingerprinting test


publicarray:
https://audiofingerprint.openwpm.com/ 🔈

CHEF-KOCH:
Bump: https://gist.github.com/CHEF-KOCH/29510713e9a2c8bd52d84ac761678a55

nodiscc:
ajax.googleapis.com can also be avoided by using https://addons.mozilla.org/en-US/firefox/addon/decentraleyes/ - https://github.com/Synzvato/decentraleyes - A Firefox add-on that emulates Content Delivery Networks (CDNs) locally by intercepting requests, finding the required resource and injecting it into the environment.

CHEF-KOCH:
Is it really necessary if we use NoScript? I doubt. Googleajax should in my opinion never blocked it breaks too much. According to the GitHub page (which I starred long time ago) it still have huge problems. Since it's not really a 'final' product I consider to not suggest it in this stage. Since it simply could more dangerous than it helps (currently). 

But I'm not into the addon to come to an final conclusion, last time I chat with the developer he said it's 'beta' and a lot is on the todo.

nodiscc:
> Is it really necessary if we use NoScript?

Nope. But in some instances when using NoScript or other Js blocking solutions, you have to whitelist a CDN when a page requires it to interact with basic functionality (for example stackexchange websites require googleapis for anything except simple logged-out browsing).

> I doubt someone really can hack or abuse any Google service since they are constantly evolved and under active development,

> I doubt someone really can hack 

Better safe than sorry. In case of a MitM attack on `ajax.googleapis.com` the attacker would have the ability to inject custom javascript in pages (_if_ the CDN has been whitelisted). Layered security. `ajax.googleapis.com` is only authenticated using TLS, which has been proved vulnerable in the past :)

I agree that Decentraleyes is still in beta. Whether we should wait to recommend it or not, I don't know. user.js already has some "experimental" (and sometimes breaking) settings, so why not. It works fine for me.

publicarray:
If you like to learn more I suggest to give the research paper by the Princeton University a read: http://randomwalker.info/publications/OpenWPM_1_million_site_tracking_measurement.pdf
or the corresponding website https://webtransparency.cs.princeton.edu/webcensus/index.html.

Ileca:
Since I am using user.js, I always wondered why it is not enforcing the strict list of the tracking protection but I guess it's because it breaks usability too much?
[...]

PR #146, opened 2016-05-21
Update README.md

CHEF-KOCH:
* Removed RC4, not supported anymore by OS or any Browser starting from oct. 2015 and higher! Google, Facebook, MS already dropped it officially (only XP is affected since no updates anymore). 
* Filldisk renamed we not need to mention the url if we anyway use url as clickable link,
* #145 added
* Browserrecon removed, useless, and obsolete since almost every test is better than this crap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,25 +283,24 @@ Online tests
 ------------
 
 * [Panopticlick](https://panopticlick.eff.org/)
-* [www.filldisk.com](http://www.filldisk.com/)
+* [Filldisk](http://www.filldisk.com/)
 * [SSL Client Test](https://www.ssllabs.com/ssltest/viewMyClient.html)
-* [evercookie](http://samy.pl/evercookie/)
+* [Evercookie](http://samy.pl/evercookie/)
 * [Mozilla Plugin Check][14]
 * [BrowserSpy.dk](http://browserspy.dk/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -283,25 +283,24 @@ Online tests
...
 * [Verify your installed Java Version](https://www.java.com/en/download/installed.jsp)
   * Protip: Don't use Oracle's Java!! But if you really need it, update it regulary!
-* [IP check](http://ip-check.info/?lang=en)
+* [IP Check](http://ip-check.info/?lang=en)
 * [Onion test for CORS and WebSocket](http://cure53.de/leak/onion.php)
 * [Firefox Addon Detector](http://thehackerblog.com/addon_scanner/)
   * [Blog post](http://thehackerblog.com/dirty-browser-enumeration-tricks-using-chrome-and-about-to-detect-firefox-plugins/)
-* [browserrecon](http://www.computec.ch/projekte/browserrecon/)??
 * [Official WebGL check](http://get.webgl.org/)
+* [AudioContext Fingerprint Test Page based on fingerprintjs2](https://audiofingerprint.openwpm.com/)
```
pyllyukko: I understood, that only parts of this is based on fingerprintjs2.

"JS/CSS and Flash font detection methods provided by fingerprintjs2 library."

PR #147, opened 2016-05-22
Update README.md - ooni-probe

CHEF-KOCH:
- added HTTP GET test which will be performed over tor or if not present an external test server
- added 'DNS related' it's important because DNS is one of the most vulnerable piece of the cake, added an test which checks if the internal DNS (OS based) is compromised to re-direct DNS requests .. Also added an suggestion to change (if in use for older FF versions) the internal DNS Cache system

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ There are several parts to all this and they are:
     - [Add-ons](#add-ons)
       - [Tracking protection](#tracking-protection)
       - [Add-ons for mobile platforms](#add-ons-for-mobile-platforms)
-  - [Online tests](#online-tests)
+  - [Test suites](#test-suites)
     - [HTML5test](#html5test)
   - [Known problems](#known-problems)
   - [CAs](#cas)
@@ -170,7 +170,6 @@ See also [#20](https://github.com/pyllyukko/user.js/issues/20).
   * Spoofs the referer header with [network.http.referer.spoofSource][9] (see: [#2](https://github.com/pyllyukko/user.js/pull/2))
   * "[Don't send the Referer header when navigating from a https site to another https site.](http://kb.mozillazine.org/Network.http.sendSecureXSiteReferrer#false)"
 * Don't accept [3rd party cookies](http://kb.mozillazine.org/Network.cookie.cookieBehavior#1)
-* An HTTP GET test can be found over [here](https://github.com/TheTorProject/ooni-probe/blob/master/docs/source/tests/http_requests.rst).
 
 ### Caching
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -170,6 +170,7 @@ See also [#20](https://github.com/pyllyukko/user.js/issues/20).
   * Spoofs the referer header with [network.http.referer.spoofSource][9] (see: [#2](https://github.com/pyllyukko/user.js/pull/2))
   * "[Don't send the Referer header when navigating from a https site to another https site.](http://kb.mozillazine.org/Network.http.sendSecureXSiteReferrer#false)"
 * Don't accept [3rd party cookies](http://kb.mozillazine.org/Network.cookie.cookieBehavior#1)
+* An HTTP GT test can be found over [here](https://github.com/TheTorProject/ooni-probe/blob/master/docs/source/tests/http_requests.rst).
```
nodiscc: s/GT/GET/

README.md
```diff
@@ -181,6 +182,10 @@ See also [#20](https://github.com/pyllyukko/user.js/issues/20).
 
 * Don't [suggest any URLs](http://kb.mozillazine.org/Browser.urlbar.maxRichResults) while typing at the address bar
 
+### DNS related
+* network.dnsCacheExpiration which shoudl be set to 0 to turn off Firefox internal DNS Cache.
```
nodiscc: s/shoudl/should

nodiscc:
How are the tests meant to be run? It seems to require ooni-probe.

CHEF-KOCH:
Of course it needs it, it's mentioned in the documentation. Any real test can't be performed by any online site because it could a) already compromised b) your system could bre already compromised. So I highly recommend to use this instead of any shitty pge which claims to 'secure' or detect some bullshit. They all not detect anything they highly depending on OS and Browser Version together with javascript, which in fact is just an low level api.

nodiscc:
Ok, then why not link directly https://github.com/TheTorProject/ooni-probe/ instead? It provides a lot of interesting tests, other than DNS and GET.

CHEF-KOCH:
Okay, updated.

nodiscc:
@CHEF-KOCH I sent a patch for your patch at https://github.com/CHEF-KOCH/user.js/pull/1. What do you think?

CHEF-KOCH:
Good.

pyllyukko:
I think we should keep the section as "Online tests". Maybe add an ```h3``` under that with "Other test suites"?

nodiscc:
My vote goes for closing this abandoned PR :/ Or apply the fix as requested above.

pyllyukko:
@CHEF-KOCH has removed the repo that had these patches. As these are pretty old indeed, it would probably need some rebasing / refactoring / further modifications (versus just hitting the "merge" button here) and that's not possible from the command line:

```
$ git checkout -b CHEF-KOCH-patch-50 master
Switched to a new branch 'CHEF-KOCH-patch-50'
$ curl -L https://github.com/pyllyukko/user.js/pull/147.patch | git am -3
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   143    0   143    0     0    131      0 --:--:--  0:00:01 --:--:--   140
100    32    0    32    0     0     15      0 --:--:--  0:00:02 --:--:--  5333
Patch is empty. Was it split wrong?
When you have resolved this problem, run "git am --continue".
If you prefer to skip this patch, run "git am --skip" instead.
To restore the original branch and stop patching, run "git am --abort".
```

Issue #75, opened 2015-11-16
include a method for users to easily include custom entries / override this user.js


berrythesoftwarecodeprogrammar:
didnt know how to word title properly.

info here  https://github.com/pyllyukko/user.js/pull/54
one may want to use this user.js but have a few things in it which they want configured differently. appending to the end of the file overrides previous entries but apparently that breaks git stuff

i just came across this http://kb.mozillazine.org/Locking_preferences
havent tried it yet but it if a file of lockprefs can override user.js then this might be a solution? people would just need to add their preferences as lockprefs in the lockprefs file. but i just realised, if it reads these files in a particular order, it may be possible to just put regular user_pref's in one of these files

will test

mengele-chan:
One possible solution is to add `override.js` and `myuser.js` to `.gitignore`. Then one can simply `git pull && cat user.js override.js > myuser.js`, no?

nodiscc:
@mengele-chan it would be a decent solution **if** it is confirmed that entries at the end of the file actually override entries above.

mengele-chan:
https://github.com/pyllyukko/user.js/pull/54#issuecomment-147024370

berrythesoftwarecodeprogrammar:
@pyllyukko ive found a solution, so you can close this if you want. maybe add some information about it to the readme for curious users?

this article has all the information: [http://kb.mozillazine.org/Locking_preferences](http://kb.mozillazine.org/Locking_preferences)

one has to create a `local-settings.js` file in their `%APP%\defaults\pref` folder. `%APP%` being the folder where the firefox or thunderbird executable resides (on windows). mine is `D:\Program Files (x86)\Mozilla Thunderbird\defaults\pref\`. the contents should be:
```
pref("general.config.obscure_value", 0);
pref("general.config.filename", "mozilla.cfg");
```

and then a `mozilla.cfg` file needs to be created in the same directory as the `.exe` e.g. `D:\Program Files (x86)\Mozilla Thunderbird\` where entries can be added to override `user.js`. they can either be normal preferences (which override), or they can be locked preferences which no addon or anything can change once the application starts. e.g:
```
lockPref("gfx.downloadable_fonts.enabled",true); // a locked preference
pref("permissions.default.image",1); // a normal preference
```

pyllyukko:
Cool, thanks. I added a note to the README.

PR #148, opened 2016-05-22
Update README.md - ciphers

CHEF-KOCH:
* You not want to hear this but it's not necessary anymore. Any related issue is already tracked down and fixed in latest Firefox version, other stuff will be reported and fixed over [here](https://github.com/mozilla/server-side-tls).

The default Ciphers in FF (today) are fine and good enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,6 @@ There are several parts to all this and they are:
     - [Caching](#caching)
     - [UI related](#ui-related)
     - [TLS / HTTPS / OCSP related](#tls--https--ocsp-related)
-    - [Ciphers](#ciphers)
   - [This is not enough!](#this-is-not-enough)
     - [Add-ons](#add-ons)
       - [Tracking protection](#tracking-protection)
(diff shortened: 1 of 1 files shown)


Gitoffthelawn:
@CHEF-KOCH Are any prefs needed to fix things if the user had created some custom prefs that were not default?

CHEF-KOCH:
Normally not, I highly suggest to reset everything and see because in meantime some Ciphers was removed due security reasons. E.g. in case you're from below 35 and want to upgrade to a recent version. Then you simply test your suite via [online test](https://www.ssllabs.com/ssltest/viewMyClient.html) [which is good enough] and then you even get warnings/suggestions if your browser use 'insecure' ones. After that you can decide to tweak it manually or not (as per needs). The theoretically attackable are maybe enabled to not break some websites. But Mozilla itself also checks certificate frauds so I would not much worry about it.

nodiscc:
@Gitoffthelawn the simplest way is to create a new profile, using the most recent user.js. In the future you can set custom preferences using [this method](https://github.com/pyllyukko/user.js/issues/75#issuecomment-204705081)

Gitoffthelawn:
@CHEF-KOCH Thanks for the details.  Good info.  Thanks for the link too.
@nodiscc Thanks for the info and the link as well!

pyllyukko:
As I [commented](https://github.com/pyllyukko/user.js/issues/158#issuecomment-223748548) in another related issue, we are not going to remove this section from the README.

PR #149, opened 2016-05-23
Welcome Chrome extensions

CHEF-KOCH:
* Added notice (ref-link) which shows how to handle Chrome extensions
* Requires FF 48 and higher

There are no about:config related things to respect but somehow I feel it needs to be mentioned and this guide is very good for beginners and answer mostly if not all questions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ It is common for [client side attacks](https://www.offensive-security.com/metasp
 * Disable Adobe Flash
 * Enable [click to play](https://wiki.mozilla.org/Firefox/Click_To_Play)
 * Enable [add-on updates](https://blog.mozilla.org/addons/how-to-turn-off-add-on-updates/)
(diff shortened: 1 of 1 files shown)


pyllyukko:
I think this is a bit irrelevant as long as we don't need / suggest using any Chrome extensions. Cool feature though.

CHEF-KOCH:
Really and what about uMatrix/uBlock or SimpleScript? You tried it? SimpleScript would be an attractive alternative to NoScript since the benefit of NoScript is more because plugins, we suggest to not use anything so I think may other Crom based on could kick in because not everyone likes NoScript and such. 

You closed this ticket way to fast. :fast_forward:

pyllyukko:
> You closed this ticket way to fast. :fast_forward:

Pull request that is. I think we don't need to have this documented at least yet. There is indeed potential in Chrome extensions, but until something concrete comes up, I think it's completely unnecessary to mention this.

PR #153, opened 2016-05-31
Suppress additional startup url

CHEF-KOCH:
First startup annoying thing. So remove it.

### DIFF
--- a/user.js
+++ b/user.js
@@ -341,6 +341,9 @@ user_pref("browser.search.suggest.enabled",		false);
 // Disable "Show search suggestions in location bar results"
 user_pref("browser.urlbar.suggest.searches",		false);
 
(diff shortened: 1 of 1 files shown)


nodiscc:
ping @pyllyukko @CHEF-KOCH

pyllyukko:
I think the first-run doesn't matter that much. Also, the new version info is actually quite nice, because at least I want to know what's changed.

pyllyukko:
I'll close this for now, as I kinda feel @CHEF-KOCH isn't changing this anymore (see my comment [here](https://github.com/pyllyukko/user.js/pull/147#issuecomment-282499092)).

Regarding your change suggestion, I think we should leave the user's homepage as it is, as it's really easy to change it through regular means and that might be something that users actually want to be able to do.

PR #154, opened 2016-05-31
Disable Dom Push

CHEF-KOCH:
Although the DOM Push API is disabled by default (probably because it requires service workers?, which are also disabled in >ESR45), set the push server URL to an empty string as a "defense in depth" measure.


pyllyukko:
Covered in #111

PR #155, opened 2016-05-31
MozTCPSocket set to false

CHEF-KOCH:
It's only relevant in ESR versions.
I never heard something is using this.


pyllyukko:
What does this setting do exactly?

PR #156, opened 2016-05-31
Disable monitoring the connected state of Tor/ESR users

CHEF-KOCH:
See [here](https://gitweb.torproject.org/user/brade/tor-browser.git/commit/?h=bug18945-01&id=d0dfca5436e051fb897f32db298093533dfe73db).

### DIFF
--- a/user.js
+++ b/user.js
@@ -120,6 +120,9 @@ user_pref("browser.fixup.alternate.enabled",		false);
 // https://trac.torproject.org/projects/tor/wiki/doc/TorifyHOWTO/WebBrowsers
 user_pref("network.proxy.socks_remote_dns",		true);
 
+// We not want to monitoring the connection state of users 
+user_pref("network.manage-offline-status",		false);
+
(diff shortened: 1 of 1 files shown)


pyllyukko:
What are the security/privacy implications of this?

I didn't see any proper reasoning in https://trac.torproject.org/projects/tor/ticket/18945

CHEF-KOCH:
> (Applies to Firefox 3.5 and above) 
Determines whether Firefox is allowed to automatically set itself to offline mode in response to certain Web sites, or if the network connection is interrupted. (See bug 620472 and this and following messages for more information.) 
True (implicit default in Firefox 3.5 and 3.6): Allows automatically setting offline mode. 
False (default in Firefox 4): Prevents automatically setting offline mode.
True in SeaMonkey 2.0 and above but SeaMonkey and Thunderbird have a separate offline manager as part of MailNews.

We not need that detection because everyone can set it manually (if necessary).

Issue #160, opened 2016-06-04
Network info fingerprinting


Roman-Nopantski:
I could find any reference to this in the repository, and noticed it wasn't already included in the user.js. Excuse my sloppy pastes with my numbering system in them :)

// 2503: disable giving away network info
   // eg bluetooth, cellular, ethernet, wifi, wimax, other, mixed, unknown, none
   // https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API
   // https://wicg.github.io/netinfo/
user_pref("dom.netinfo.enabled", false);

CHEF-KOCH:
None of the dom.* settings can compromise any security aspect. Why the recent dom.* stuff was merged is beyond me. 

This is something you can be enabled, because it requires other services instaled + activated which directly using it. Maybe WhatsApp, or Telegram (WebChat) need this but I would not disable it because this not compromises something, or in other words it's to useful that someone need to disable it.

Roman-Nopantski:
Totally agree that it is harmless. Its not a security issue at all (AFAIK). It's fingerprinting issue - hence the title :) It's just a suggestion.

Maybe I'm confused. I know this user.js is about "hardening" Firefox. You clearly look at security issues. You also have some tracking and privacy settings, and I bet if I look, there are fingerprinting ones. I do realize there can be overlap between these things.

OK, I just had a quick check. Yes, you have lots of fingerprinting settings - such as Battery API, graphics hardware ( gfx.direct2d.disabled etc), media device enumeration ( media.navigator.enabled ), gamepad API which leaks USB device IDs, Virtual Reality settings, media JS performance fingerprinting ( media.video_stats.enabled ) ... to name a few.

PS: There's also two harmless geo* settings.

PPS: Not sure what you mean that it "requires other services installed + activated which directly using it". Its an API which any script can call - right?

CHEF-KOCH:
Entirely disable dom and every 'logger' breaks a lot of stuff e.g. uBlocks Dashboard/Waterfile logger because such addons are highly depending on DOM. I not know that DOM was compromised or fingerprinted directly since it's executed locally and not like the other stuff online, this is a difference. 

If you found something which is useless or harmless feel free to discuss this in the pull request or discussion topic. 

[...]

PR #161, opened 2016-06-04
Disable dom.netinfo.enabled - Network info fingerprinting

CHEF-KOCH:
Merge suggestion from #160.


Atavic:
[A request](https://github.com/pyllyukko/user.js/issues/160#issuecomment-227071498) can be satisfied looking at the Connection Type result on http://browserspy.dk/offline.php

[here](http://wicg.github.io/netinfo/) are the available entries.

PR #175, opened 2016-08-31
add OS X paths

publicarray:
For reference: https://developer.mozilla.org/en-US/Firefox/Enterprise_deployment#Mac_Directory_Change


pyllyukko:
Thanks!

PR #180, opened 2016-09-14
Remove obsolete browser.frames.enabled pref

fmarier:
This was removed in https://bugzilla.mozilla.org/show_bug.cgi?id=1013457


pyllyukko:
Thanks!

PR #183, opened 2016-09-21
Don't make WebGL debug info available to websites

fmarier:
This is currently only enabled on Nightly and Aurora and only
applies to people who choose to comment out the webgl.disabled line.


pyllyukko:
Thanks!

PR #184, opened 2016-09-23
Update the WebRTC prefs for Firefox 51

fmarier:
See discussion on https://bugzilla.mozilla.org/show_bug.cgi?id=1304600.

### DIFF
--- a/user.js
+++ b/user.js
@@ -40,6 +40,7 @@ user_pref("dom.netinfo.enabled",				false);
 // https://wiki.mozilla.org/Media/WebRTC/Privacy
 user_pref("media.peerconnection.ice.default_address_only",	true); // Firefox < 51
 user_pref("media.peerconnection.ice.no_host",			true); // Firefox >= 51
+// Disable WebRTC entirely
 user_pref("media.peerconnection.enabled",			false);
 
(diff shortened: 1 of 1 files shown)


user.js
```diff
@@ -38,7 +38,8 @@ user_pref("dom.netinfo.enabled",				false);
 // Don't reveal your internal IP
 // Check the settings with: http://net.ipcalf.com/
 // https://wiki.mozilla.org/Media/WebRTC/Privacy
-user_pref("media.peerconnection.ice.default_address_only",	true);
+user_pref("media.peerconnection.ice.default_address_only",	true); // Firefox < 51
+user_pref("media.peerconnection.ice.no_host",			true); // Firefox >= 51
 user_pref("media.peerconnection.enabled",			false);
```
ivan-kolmychek: I think it would be nice to also have the comment here explaining what does setting `media.peerconnection.enabled` to `false` does. From how I understand it, it disables the WebRTC entirely, but I'm not sure about that.

fmarier: Good idea. I've added that in 9a7a697d03f9561227c82657e05de3013804f169.

pyllyukko:
Thanks!

PR #187, opened 2016-10-04
Recommend Smart HTTPS

fmarier:
This add-on changes the default (from HTTP to HTTPS) for URLs that you type into the address bar.

I recommend clicking the "incognito" setting after installation.

It seems like a nice addition to the list of suggested add-ons since it covers sites that HTTPS Everywhere doesn't know about yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,7 @@ Here is a list of the most essential security and privacy enhancing add-ons that
 * [Certificate Patrol][4]
   * I recommend setting the 'Store certificates even when in [Private Browsing][8] Mode' to get full benefit out of certpatrol, even though it stores information about the sites you visit
 * [HTTPS Everywhere](https://www.eff.org/https-everywhere)
+* [Smart HTTPS](https://addons.mozilla.org/firefox/addon/smart-https/)
 * [NoScript](http://noscript.net/)
 * [DuckDuckGo Plus](https://addons.mozilla.org/firefox/addon/duckduckgo-for-firefox/) (instead of Google)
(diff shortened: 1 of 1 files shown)


ivan-kolmychek:
Is its source code available? What is its license? I've not found the mention of this on "official" page.

Gitoffthelawn:
@ivan-kolmychek Source code is available here: https://addons.mozilla.org/firefox/files/browse/513702/

nodiscc:
So does it force HTTPS? Because it _sometimes_ breaks websites.

fmarier:
> So does it force HTTPS?

It does, but it allows you to switch back to HTTP and whitelist "broken" sites manually.

nodiscc:
See also https://github.com/pyllyukko/user.js/issues/188

pyllyukko:
> It seems like a nice addition to the list of suggested add-ons since it covers sites that HTTPS Everywhere doesn't know about yet.

HTTPS Everywhere's whitelisting method indeed isn't perfect, but it feeld kinda redundant to have several add-on recommendations on the same topic.

ivan-kolmychek:
@pyllyukko well, we can suggest to choose one of them and stick with it, but providing alternatives is not a bad idea, IMHO. :)

ivan-kolmychek:
@nodiscc https everywhere can force HTTPS too - see "Block all unencrypted requests" option in menu.

pyllyukko:
> @pyllyukko well, we can suggest to choose one of them and stick with it, but providing alternatives is not a bad idea, IMHO. :)

Well that's true also. Should we still have this for instance after the HTTPS Everywhere in parentheses, e.g. "HTTPS Everywhere (or Smart HTTPS)". Because there's probably no point in having both and someone might interpret it as such.

PR #190, opened 2016-10-12
Remove network.http.sendSecureXSiteReferrer

fmarier:
This pref is unnecessary if you spoof or turn of referrers. Also, it makes no sense to restrict HTTPS referrers while allowing HTTP referrers.

It has been removed from Firefox 52 in https://bugzilla.mozilla.org/show_bug.cgi?id=1308725.


pyllyukko:
True that. Thanks.

PR #191, opened 2016-10-14
Replace "Smart HTTPS" with "HTTPS by default"

fmarier:
[Smart HTTPS](https://addons.mozilla.org/firefox/addon/smart-https) seems to break a few sites in non-obvious ways and its defaults are not great. [HTTPS by default](https://addons.mozilla.org/firefox/addon/https-by-default) has none of these problems and just does one thing well: it changes the default scheme when typing scheme-less URLs in the address bar.

Also add the trimURLs setting to user.js since it's [recommended by the author of "HTTPS by default"](https://github.com/Rob--W/https-by-default#firefox) and it will probably appeal to the user.js crowd too.


Atavic:
Electrolysis compatible.

Issue #188, opened 2016-10-05
[Addon] HTTP Nowhere


nodiscc:
Would https://addons.mozilla.org/en-US/firefox/addon/http-nowhere/ be worth mentioning on the README? It blocks _all_ connections not made over HTTPS. There is also an option to try forcing HTTPS when the site presents itself over a plaintext connection (similar to SMart HTTPS https://github.com/pyllyukko/user.js/pull/187#issuecomment-251640511), and a manual whitelist.

It _does_ break a lot of websites which is expected since many hosts are not configured for SSL/TLS.  
It does exactly what it says, block HTTP traffic.

Only major problem I found: it blocks connections to OCSP servers, which in turn also causes HTTPS connections to error. You have to manually whitelist them. In that regard https://github.com/pyllyukko/user.js/issues/73 could be useful, I did not go this far.

https://github.com/cwilper/http-nowhere/ has some info.

Atavic:
[This](https://github.com/pyllyukko/user.js/issues/73) whitelist seems a must for this addon to work properly. Without OCSP renewal, the browser will end finding expired CA Certs.

Another way is [here](https://github.com/cwilper/http-nowhere/issues/16).

pyllyukko:
> It does break a lot of websites

If this is the case, I think we shouldn't recommend it. I'm sure the world isn't ready for this yet :)

Atavic:
I'd like to post a recent problem on Soundcloud: `Vendor Downtime Causing Certificate Messages`
caused by [GlobalSign experiencing issues with OCSP](https://twitter.com/globalsign/status/786505261842247680) and offering [this solution](https://support.globalsign.com/customer/portal/articles/1353318) because _Invalid certificate remains in cache_

ivan-kolmychek:
`HTTP Nowhere` haven't been updated for a while already (`Last Updated: September 16, 2013`) and HTTPS Everywhere does have the support of blocking non-encrypted communications. 

Maybe, that feature should be mentioned in readme instead of adding the notion of yet another addon? Especially given that HTTPS Everywhere is *already* recommended there. :)

![screenshot_20161016_184613](https://cloud.githubusercontent.com/assets/5011577/19418678/7773220a-93d1-11e6-8cf8-ff2f552b9e4d.png)

ivan-kolmychek:
On the second thought, I haven't found the manual whitelisting in HTTPS Everywhere, that's my bad.

Atavic:
[...]

PR #195, opened 2016-10-26
doc: add note about blocking all unencrypted requests

nodiscc:
Fixes https://github.com/pyllyukko/user.js/issues/188


pyllyukko:
Should we add some disclaimer, that this will probably break the internets for the user? :) As there are still sites/servers without HTTPS.

nodiscc:
Updated.

PR #196, opened 2016-10-27
Update value of security.pki.sha1_enforcement_level for Firefox >= 52

fmarier:
https://bugzilla.mozilla.org/show_bug.cgi?id=1302140
https://dxr.mozilla.org/mozilla-central/rev/f9f3cc95d7282f1fd83f66dd74acbcdbfe821915/security/certverifier/CertVerifier.h#143-152

### DIFF
--- a/user.js
+++ b/user.js
@@ -610,9 +610,9 @@ user_pref("security.tls.version.max",				3);
 // "2. Strict. Pinning is always enforced."
 user_pref("security.cert_pinning.enforcement_level",		2);
 
-// allow SHA-1 only if notBefore < 2016-01-01
-// https://bugzilla.mozilla.org/show_bug.cgi?id=942515#c32
-user_pref("security.pki.sha1_enforcement_level",		2);
+// disallow SHA-1
+// https://bugzilla.mozilla.org/show_bug.cgi?id=1302140
+//user_pref("security.pki.sha1_enforcement_level",		1);
 
 // https://wiki.mozilla.org/Security:Renegotiation#security.ssl.treat_unsafe_negotiation_as_broken
 // see also CVE-2009-3555


pyllyukko:
Why we should allow it for custom CAs?

fmarier:
> Why we should allow it for custom CAs?

I don't know why anybody using `user.js` would install a custom CA (other than temporarily for debugging webapps maybe), but if they did and it used SHA1, then every single HTTPS site would be broken.

Also, if you're using such a local man-in-the-middle proxy (e.g. anti-virus), then there's no network traffic anyways so the strength of the encryption doesn't matter at all.

fmarier:
Since Firefox 52 [defaults to `4`](http://searchfox.org/mozilla-central/rev/445097654c1cc1098ee0171a29c439afe363a588/netwerk/base/security-prefs.js#59), it becomes a little tricky to set this properly because:

- If we default to `2`, Firefox >= 52 will see a fair amount of breakage since it's an alias for `1`.
- If we default to `4` on Firefox >= 52, we will not benefit from the [default changing to `3`](https://blog.mozilla.org/security/2016/10/18/phasing-out-sha-1-on-the-public-web/) in 2017 unless we update again.

I'm tempted to replace all of this with a commented-out `security.pki.sha1_enforcement_level = 1` so that those people who want to fully disable it can and the rest will get the defaults which are very good starting on Firefox 52. Does that sound like a good idea?

Roman-Nopantski:
Yes - sounds like the ONLY idea - been running it at 1 for a year with no problems (I do not have any MitM devices such as AV interfering with my web traffic). I am but one user, but the ghacks user.js which I authored is used by tons of people as a template - no-one has complained about it on there (mind you, I do provide "warning" tags and a troubleshooting section). If users are knowledgeable enough to use a user.js then they can figure it out.

pyllyukko:
Yes, this sounds like the best approach. The at some point uncomment it, when the world is ready.

fmarier:
Patch updated.

pyllyukko:
SHA-1 collision:
  * <http://shattered.io/>
  * <https://security.googleblog.com/2017/02/announcing-first-sha1-collision.html>

pyllyukko:
I'm thinking of uncommenting this so the value would be ```1```. Any objections?

pyllyukko:
Done: eb4e5ddbcbb5efb1284ac17ad9528f94007e67b2

Roman-Nopantski:
Wow .. 8 minutes between asking and committing, excellent :) FYI, I have been using the equivalent of 1 for IDK, a year, year and a half, maybe 2 years since I started this thing. Zero impact, and zero complaints.

pyllyukko:
It was almost 9 minutes :)

Cool. Same here and haven't noticed anything. Good riddance!

Roman-Nopantski:
> It was almost 9 minutes :)

Don't blame me bro, blame github's minute rounding xD

PR #201, opened 2017-01-06
More HTTPS

devmapper0:
There is so quiet last time, so I have changed some links to HTTPS. Changed URL where I was redirected. "Are we... something" from Ghostery is down, so I added archive.org link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ This is a [user.js][1] configuration file for Mozilla Firefox that's supposed to
 ### Main goals
 
 * Limit the possibilities to track the user through [web analytics](https://en.wikipedia.org/wiki/Web_analytics)
-* Harden the browser, so it doesn't spill its guts when asked (have you seen what [BeEF](http://beefproject.com/) can do?)
+* Harden the browser, so it doesn't spill its guts when asked (have you seen what [BeEF](https://beefproject.com/) can do?)
 * Limit the browser from storing anything even remotely sensitive persistently (mostly just making sure [private browsing][8] is always on)
 * Make sure the browser doesn't reveal too much information to [shoulder surfers](https://en.wikipedia.org/wiki/Shoulder_surfing_%28computer_security%29)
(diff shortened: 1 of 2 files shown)


pyllyukko:
Thanks! :)

Indeed it has been quiet recently. But we're still here!

Roman-Nopantski:
line 353: [Test page for Firefox's built-in Phishing Protection](https://itisatrap.org/firefox/its-a-trap.html) ("Web forgeries") = 403 Forbidden
Correct url is https://www.itisatrap.org/firefox/its-a-trap.html
--
line 355: [Test page for Firefox's built-in Malware Protection](https://itisatrap.org/firefox/unwanted.html) (unwanted software) = 403 Forbidden
Correct url is https://www.itisatrap.org/firefox/unwanted.html

pyllyukko:
Hmm. Both URLs work for me without the ```www.``` subdomain. But the <https://itisatrap.org/firefox/its-a-tracker.html> one doesn't.

fmarier:
The URLs should work fine without the `www`. However, it looks like these pages all broke recently.

I filed https://github.com/mozilla/itisatrap/issues/19 to track this.

Roman-Nopantski:
I added these about 4 or five months ago (not sure if you have them all)
   Attack            https://itisatrap.org/firefox/its-an-attack.html
   Blocked           https://itisatrap.org/firefox/blocked.html
   Malware           https://www.itisatrap.org/firefox/unwanted.html
   Phishing          https://www.itisatrap.org/firefox/its-a-trap.html
   Tracking          https://itisatrap.org/firefox/its-a-tracker.html
They weren't ALL HTTPS, so I changed a couple, and then, TBH, my FF is so screwed down, I just tested the URLs in IE. I will fire up a vanilla portable FF and test from there and report back :)

Roman-Nopantski:
@fmarier : I just ran a vanilla FF 50.1 and loaded all five urls with and without the www, so 10 pages in all. All five tests without a www returned a 403, all urls with a www were fine. You might want to log some more tickets

PR #206, opened 2017-02-15
Bump max TLS version

DrunkenSasquatch:
`security.tls.version.max=4` is compatible with Firefox versions 51.0.1 and above.


Roman-Nopantski:
IIRC, this broke in 51 (and was fixed in 51.01) when they fixed `user_pref("security.tls.version.fallback-limit", 3);` , so if you're going to include it, I suggest you also include this one for reference

DrunkenSasquatch:
@Roman-Nopantski Thanks

pyllyukko:
Thanks!

Issue #208, opened 2017-02-20
ghacks user.js monster diff


pyllyukko:
Based on @Roman-Nopantski's diff: <https://gist.github.com/pyllyukko/f5184fbb51b5e340f5637adee582c4d9>

STARTUP
-------

* [x] 0101: disable "slow startup" options
  * WONTFIX: Doesn't seem that relevant
* [x] 0102: set start page (0=blank, 1=home, 2=last visited page, 3=resume previous session)
  * Commented out in the ghacks version
  * Some aspects considered in #218

GEOLOCATION
-----------

* [x] 0201: disable location-aware browsing
  * Should be covered by ```geo.enabled```
  * Also Mozilla bug's [689252](https://bugzilla.mozilla.org/show_bug.cgi?id=689252) & [692927](https://bugzilla.mozilla.org/show_bug.cgi?id=692927) talk that ```geo.wifi.*``` settings are not used anymore.
* [x] 0202: disable GeoIP-based search results
  * 00b102e9a2f6e6c615da8a96c67c47d0af7b9a59
* [ ] 0203: disable using OS locale, force APP locale
* [x] 0204: set APP local
* [x] 0206: disable geographically specific results/search engines eg: "browser.search.*.US"
* [x] 0207: set language to match
  * d80e67469d0219b46836f737441006512febcf0e
* [x] 0208: enforce US English locale regardless of the system locale

QUIET FOX [PART 1]
------------------

* [x] 0301: disable browser auto update
  * WONTFIX: Updates are good for you :)
* [x] 0305: disable add-ons auto update
  * WONTFIX: Updates are good for you :)
* [x] 0307: disable auto updating of personas (themes)
* [x] 0309: disable sending Flash crash reports
* [x] 0310: disable sending the URL of the website where a plugin crashed
  * Need more info on ```dom.ipc.plugins.reportCrashURL```
* [x] 0320: disable extension discovery (featured extensions)
  * WONTFIX
* [x] 0330b: set unifiedIsOptIn to make sure telemetry respects OptIn choice and that telemetry
  * WONTFIX: Telemetry is already disabled
* [x] 0331: remove url of server telemetry pings are sent to
  * WONTFIX: Telemetry is already disabled
* [x] 0332: disable archiving pings locally - irrelevant if toolkit.telemetry.unified is false
  * f781e7d27aa77d5fe2bf27a81cf966ad029280b7
* [x] 0333a: disable health report
[...]

PR #214, opened 2017-02-25
disk cache: add 2 commented out prefs for reference

nodiscc:
Fixes item 1001 in #208

### DIFF
--- a/user.js
+++ b/user.js
@@ -500,7 +500,7 @@ user_pref("places.history.enabled",				false);
 user_pref("network.cookie.lifetimePolicy",			2);
 
 // http://kb.mozillazine.org/Browser.cache.disk.enable
-user_pref("browser.cache.disk.enable",				false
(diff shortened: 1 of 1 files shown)


pyllyukko:
I think this needs some additional researching and links in the user.js. Sounds like something that ```browser.cache.disk.enable``` already handles.

nodiscc:
Added documentation

pyllyukko:
Still... do we need these at all? Smart or dumb, the disk cache isn't in use and that can be verified from ```about:cache```:

```
Number of entries: 	0
Maximum storage size: 	32768 KiB
Storage in use: 	0 KiB
Storage disk location: 	none, only stored in memory 
```
Of course there's the "When disk cache is enabled" note, but is it unnecessary clutter?

nodiscc:
You are right. Closing, feel free to mark #208 item 1001 as fixed (#214)

pyllyukko:
Done.

PR #215, opened 2017-02-25
Set security.password_lifetime = 5

nodiscc:
Fixes item 0904 in #208


Roman-Nopantski:
That needs some qualification: `security.ask_for_password` must be -> 2

/* 0903: set how often Mozilla should ask for the master password
 * 0=the first time (default), 1=every time it's needed, 2=every n minutes (as per the next pref) ***/
user_pref("security.ask_for_password", 2);

pyllyukko:
Thanks for the clarification.

PR #217, opened 2017-02-25
doc: add CVE references for ServiceWorkers

nodiscc:
Details on each CVE can be found at Mozilla Security Advisories, CVEDetails or NIST.
Known vulnerabilities seem to be fixed.
Ref. https://github.com/pyllyukko/user.js/commit/c65632f066223516d26ea67aecfa7a9fb04a3238


pyllyukko:
This PR also reverts the stuff from #216. I'm not entirely sure what kind of Git-fu it would require for me to fix that.

pyllyukko:
I ```cherry-pick```ed the commit to fix this. I think this will soon turn into a guitar hero, but that's ok as it's good Git-fu practice :)

guitar_hero

PR #218, opened 2017-02-25
document startup.homepage* prefs

nodiscc:
Add disabled prefs to UI section
May be useful as some users might consider these "automatic connections"
Ref. #153

### DIFF
--- a/user.js
+++ b/user.js
@@ -617,6 +617,18 @@ user_pref("security.ask_for_password",				0);
 // If password manager/master password is enabled, lock password storage every 5 minutes (default is 30)
 user_pref("security.password_lifetime", 5);
 
+// Set homepage at browser startup to empty page
+// user_pref("browser.startup.homepage", "about:blank");
+
+// Don't show information page after upgrades
+// http://kb.mozillazine.org/Startup.homepage_override_url
+// user_pref("startup.homepage_override_url", "about:blank");
+
+// Disable first-run welcome page
+// http://kb.mozillazine.org/Startup.homepage_welcome_url
+// user_pref("startup.homepage_welcome_url", "about:blank");
+// user_pref("startup.homepage_welcome_url.additional", "about:blank");
+
 /******************************************************************************
  * TLS / HTTPS / OCSP related stuff                                           *
  *                                                                            *


user.js
```diff
@@ -178,6 +178,10 @@ user_pref("security.mixed_content.block_active_content",	true);
 // Mixed Passive Content (a.k.a. Mixed Display Content).
 user_pref("security.mixed_content.block_display_content",	true);
 
+// https://secure.wikimedia.org/wikibooks/en/wiki/Grsecurity/Application-specific_Settings#Firefox_.28or_Iceweasel_in_Debian.29
+user_pref("javascript.options.methodjit.chrome",		false);
+user_pref("javascript.options.methodjit.content",		false);
+
```
pyllyukko: What should we do with these?

pyllyukko: What should we do with these?

pyllyukko:
Should you ```add -p``` these so that you don't revert previous stuff?

nodiscc:
Rebased

pyllyukko:
I still feel that these homepage thingies aren't necessary in the user.js. Might be minor annoyance, but not really any kind of privacy/security issue. Wouldn't you say?

And also to keep it tidier.

pyllyukko:
Maybe we could document these in the wiki?

nodiscc:
> I still feel that these homepage thingies aren't necessary in the user.js. Might be minor annoyance, but not really any kind of privacy/security issue. Wouldn't you say?

I agree. Let's close this PR.

> Maybe we could document these in the wiki?

:/ I think information is currently too much scattered between user.js comments, the wiki, the readme... In my experience it leads to duplicate/outdated/don't-know-where-to-look documentation. It can be made cleaner eg by:

 * Moving all info from these wiki pages to the user.js file itself, alongside relevant prefs/sections
   - [ ] https://github.com/pyllyukko/user.js/wiki/Home
   - [ ] https://github.com/pyllyukko/user.js/wiki/Caching
   - [ ] https://github.com/pyllyukko/user.js/wiki/CIPHERS
   - [ ] https://github.com/pyllyukko/user.js/wiki/Extensions-&-Plugins
   - [ ] https://github.com/pyllyukko/user.js/wiki/Firefox-exploits-&-vulnerabilities
   - [ ] https://github.com/pyllyukko/user.js/wiki/Firefox-features-and-components
   - [ ] https://github.com/pyllyukko/user.js/wiki/HTML5---APIs---DOM
   - [ ] https://github.com/pyllyukko/user.js/wiki/HTTP
   - [ ] https://github.com/pyllyukko/user.js/wiki/Local-Dom-Storage
   - [ ] https://github.com/pyllyukko/user.js/wiki/Misc
   - [ ] https://github.com/pyllyukko/user.js/wiki/TLS---HTTPS----OCSP
   - [ ] https://github.com/pyllyukko/user.js/wiki/UI-related
   - [ ] https://github.com/pyllyukko/user.js/wiki/Firefox-autoconnections
   * There's at most a few links by section/page
   * Wiki tables appear to be broken
   * If the info does not fit in `user.js` it's probably not relevant, remove it.
* Move some tests from https://github.com/pyllyukko/user.js#online-tests to relevant prefs/sections in `user.js`
* Move info from https://github.com/pyllyukko/user.js#what-does-it-do to section descriptions in `user.js`, and instead generate the "What does it do?" section do from section descriptions in `user.js` (ezpz sed/awk/Makefile magic)
* Move https://github.com/pyllyukko/user.js/wiki/FAQ,  to README.md

So closing. Let me know if you agree with the above and if I should open a new issue.

pyllyukko:
> Moving all info from these wiki pages to the user.js file itself, alongside relevant prefs/sections

Yes, the wiki should be cleaned up, gathering the relevant parts and then closing it completely. All the relevant info should be in the README.

>         If the info does not fit in user.js it's probably not relevant, remove it.

I agree.

>     Move some tests from https://github.com/pyllyukko/user.js#online-tests to relevant prefs/sections in user.js

Moved or copied? I think the online tests section is good in the README as it is, but some of the tests could be also in the user.js.

>     Move info from https://github.com/pyllyukko/user.js#what-does-it-do to section descriptions in user.js, and instead generate the "What does it do?" section do from section descriptions in user.js (ezpz sed/awk/Makefile magic)

I don't think that's a good idea. I think we should have it the similar way it's now, kinda summarizing different aspects of the hardening.

>     Move https://github.com/pyllyukko/user.js/wiki/FAQ, to README.md

Haven't looked what's there :P

>     Closing the wiki: Github Flavored Markdown is not very portable, there are authenticity issues with world-writable wikis, all in all it's harder to maintain than generating it from static comments/.md files. If you think this bloats the user.js file/README, just discard irrelevant information.

Yes.

PR #239, opened 2017-03-14
Fix open issues/PRs #197 #108 #111 #236 #238

nodiscc:
~~Incomplete PR in the current state (#238 work still going on) but the~~ first commits [can be reviewed/merged independently](https://gist.github.com/piscisaureus/3342247) (see commit messages):

* add firefox version indication for battery api (fixes #197)
* add commented out prefs for *effective* default search engine choice (closes #108)
* document disabling push notifications (references in commit message, closes #111)
* Add missing preferences descriptions (Fixes #236)
 * set security.ask_for_password to 2 (lock password storage periodically), set password storage lock timeout to 1 minute (fixes #235)
* improve and reorder FAQ 			d1e4b6d
* rename LICENSE.txt file to LICENSE, link to it from README 		eb45fd7
* more explicit NOTICE: fields 			581309b
* gen-readme.sh: add 'known problems and limitations' section generation 			930967b
* run make, generate 'kwown problems and limitations' readme section, fixes #238  b70f602

PR #242, opened 2017-03-14
README: separate CAs hardening documentation

nodiscc:
This is based on #239 and adds a single commit/change:

* split CAs hardening guide to separate file (**greatly** improves readability of README),
* Increase CA problem visibility and link to it from 'Further hardening' section
* deduplicate usage information, link to appropriate sections instead of repeating ourselves

-------------------------------------------------------

I believe this is a much needed change to get users/contributors to

 * actually read the whole readme
 * actually install user.js without getting rebutted by the apparent complexity of the procedure
 * seriously consider trimming down CA lists as a further hardening step

Preview at:

https://github.com/nodiscc/user.js/blob/readme-separate-cas-doc/README.md#further-hardening
https://github.com/nodiscc/user.js/blob/readme-separate-cas-doc/CAs.md

### DIFF
--- a/CAs.md
+++ b/CAs.md
@@ -1,0 +1,95 @@
+# Limiting trust in Certificate Authorities
+
+
+It all started when I read [this blog post][5]...
+
+So another part of my browser hardening was to somehow reduce the number of CAs trusted by my browser. First I thought I would sniff all the HTTPS connections and extract the certificates from there, to get the list of CAs I **really** need.
+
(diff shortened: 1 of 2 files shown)


pyllyukko:
I agree that the CAs section might be a bit overkill for the main README. Also looking at the issues/PRs, there doesn't seem to be much attention against it.

It's still an important part of hardening the browser, but it might be better to separate it.

So apparently we should merge either this or #239?

nodiscc:
> So apparently we should merge either this or #239?

We should wait for #239 to be merged, then merge this (or `git cherry-pick` https://github.com/pyllyukko/user.js/pull/242/commits/3974a01df22f45ad2d1ebe5d26801c7a50eac897 but I don't think it can be applied on top of current master that easily, git will probably complain and request manual editing)

nodiscc:
Rebased.

Issue #86, opened 2015-11-24
Consider enabling WebRTC in the future


pyllyukko:
See #84.

nodiscc:
Firefox 52 is the new [ESR](https://www.mozilla.org/en-US/firefox/organizations/all/) - in which the LAN IP address leak is fixed. So we should either:

 * comment out the pref disabling WebRTC (`media.peerconnection.enabled`)
 * leave it as is for hardening purposes, and close this issue
  * if notable functionality breakage is found due to WebRTC being disabled, add a `NOTICE:` field alongside that pref.

fmarier:
`media.peerconnection.ice.default_address_only` should also be removed since the way to disable the leaking of internal IP addresses is `media.peerconnection.ice.no_host` in Firefox >= 52.

nodiscc:
@fmarier `media.peerconnection.ice.no_host` is already present (https://github.com/pyllyukko/user.js/blob/master/user.js#L61)

`media.peerconnection.ice.default_address_only` is there for legacy reasons (with a clear indicator that this is for firefox < 51).

Or should we remove _all_ prefs for versions older than the current ESR? (whole other issue)

pyllyukko:
> @fmarier media.peerconnection.ice.no_host is already present (https://github.com/pyllyukko/user.js/blob/master/user.js#L61)

The comment in user.js says ```Firefox >= 51``` and @fmarier said it's ```Firefox >= 52```...

> media.peerconnection.ice.default_address_only is there for legacy reasons (with a clear indicator that this is for firefox < 51).

Yep.

> Or should we remove all prefs for versions older than the current ESR? (whole other issue)

No. At least not yet, as ESR 52 is just out of the oven and some distros are still using ESR 45.*.

nodiscc:
> No. At least not yet, as ESR 52 is just out of the oven and some distros are still using ESR 45.*.

https://github.com/pyllyukko/user.js/issues/251

> The comment in user.js says Firefox >= 51 and @fmarier said it's Firefox >= 52...

Yes, this should be clarified. @fmarier is probably right, will edit the comment.

I will amend https://github.com/pyllyukko/user.js/pull/249 to re-disable WebRTC. If there are not other points to discuss I think this issue can be closed (changes like this can be tracked in #251)

nodiscc:
[...]

PR #250, opened 2017-03-20
Update name of pref for phishing protection

fmarier:
This pref got [renamed in Firefox 50](https://bugzilla.mozilla.org/show_bug.cgi?id=1025965).

### DIFF
--- a/user.js
+++ b/user.js
@@ -400,12 +400,13 @@ user_pref("loop.logDomains",					false);
 //user_pref("app.update.auto",		true);
 
 // PREF: Enable blocking reported web forgeries
-// http://kb.mozillazine.org/Browser.safebrowsing.enabled
+// https://wiki.mozilla.org/Security/Safe_Browsing
 // http://kb.mozillazine.org/Safe_browsing
 // https://support.mozilla.org/en-US/kb/how-does-phishing-and-malware-protection-work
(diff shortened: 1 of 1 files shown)


user.js
```diff
@@ -400,12 +400,12 @@ user_pref("loop.logDomains",					false);
...
 
 // PREF: Enable blocking reported web forgeries
-// http://kb.mozillazine.org/Browser.safebrowsing.enabled
+// https://wiki.mozilla.org/Security/Safe_Browsing
 // http://kb.mozillazine.org/Safe_browsing
 // https://support.mozilla.org/en-US/kb/how-does-phishing-and-malware-protection-work
 // http://forums.mozillazine.org/viewtopic.php?f=39&t=2711237&p=12896849#p12896849
 // CIS 2.3.4
-user_pref("browser.safebrowsing.enabled",			true);
+user_pref("browser.safebrowsing.phishing.enabled",		true);
```
nodiscc: Perhaps the 2 prefs should be kept with a comment noting which FF version is affected? (At least while we still supprt FF versions >= 45, see https://github.com/pyllyukko/user.js/issues/86#issuecomment-287873061):

```
user_pref("browser.safebrowsing.enabled",			true);  // (<50)
user_pref("browser.safebrowsing.phishing.enabled",		true); // (>=50)
```

fmarier: Updated.

PR #255, opened 2017-03-25
README: refactor addons section; other fixes; fix GEOLOCATION/QUIET FOX sections of #208

### LINKED ISSUE #33: Consider removing old settings

### LINKED ISSUE #193: Discussion: UA: window vs iframe
see: https://github.com/gorhill/uMatrix/issues/649

Personally, I do not try to hide my browser as being FF, but I do always spoof it as the latest ESR. As crssi says in the above link, if you go to https://www.browserleaks.com/javascript and check your UA values in both `window` and `iframe.contentWindow` , if you are spoofing, you will get different results.
- crssi uses uMatrix to spoof
- I use UAControl and User-Agent JS-Fixer

Does anyone else get this behavior with other extensions? Is this a bug that someone can let Mozilla know about?

Edit: Assuming the user spoofs the same value for both headers and JS. The iframe leaks the real value.

nodiscc:
See commit messages:

```
* c850e47 README: add-ons: add extra blocklists recommendation, shorten addon-specific recommendations
* 97b3166 README: refactor addons section:  * remove lengthy description, just recommend addons, or don't recommend them.  * remove duckduckgo plus, add generic hardening recommendation about privacy-conscious search engines  * store all addon-specific recommendations in the addon section  * Move ghostery to Other addons subsection, add uMatrix to this section  * remove links to issues (use the search feature) * move built-in tracking protection links to user.js tracking protection section * run make - auto-update 'What does it do' section
* 311c51d README: improve FAQ about deprecated preferences, closes #33 #251
* cdf0bf5 simplify privacy.resistfingerprinting references: link to mozilla meta tracking page for this pref change pref description since this is related to more than screen size and media queries
* 47bbbe5 improve comment-out referer spoofing prefs, closes #193
```

See https://github.com/nodiscc/user.js/tree/fixesfixesfixes#add-ons

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,16 @@
+pyllyukko <pyllyukko@maimed.org>
+nodiscc <nodiscc@gmail.com>
+CHEF-KOCH <CHEF-KOCH@users.noreply.github.com>
+Francois Marier <francois@mozilla.com>
+CHEF-KOCH <Nvinside@gmail.com>
+Francois Marier <francois@debian.org>
+Alex Hirsch <w4rh4wk@bluephoenix.at>
+DrunkenSasquatch <j.hamshere91@gmail.com>
+Frank LENORMAND <lenormf@gmail.com>
+Sebastian Schmidt <publicarray@icloud.com>
+Steffen Gransow <graste@mivesto.de>
+devmapper0 <devmapper0@users.noreply.github.com>
+Mehmet Atif Ergun <mehmetaergun@users.noreply.github.com>
+mengele-chan <mengele-chan@yandex.ru>
+uberspot <onexemailx@gmail.com>
+zummuz <zummuz@users.noreply.github.com>
--- a/Makefile
+++ b/Makefile
@@ -1,2 +1,56 @@
+all: whatdoesitdo tests authors
+
 whatdoesitdo:
+	@# generate the README "What does it do?" section
 	@./gen-readme.sh
+
+# To decrease tests verbosity, comment out unneeded targets
+tests: downloadffprefs checknotcovered checkdeprecated stats cleanup
+
+
+downloadffprefs:
+	@# download and sort all known preferences files from Firefox (mozilla-central) source
+	@# specify wanted Firefox version/revision below (eg. "tip", "FIREFOX_AURORA_45_BASE", "9577ddeaafd85554c2a855f385a87472a089d5c0"). See https://hg.mozilla.org/mozilla-central/tags
+	@SOURCEVERSION="tip"; \
+	FIREFOX_SOURCE_PREFS=" \
+	https://hg.mozilla.org/mozilla-central/raw-file/$$SOURCEVERSION/toolkit/components/telemetry/datareporting-prefs.js \
+	https://hg.mozilla.org/mozilla-central/raw-file/$$SOURCEVERSION/toolkit/components/telemetry/healthreport-prefs.js \
+	https://hg.mozilla.org/mozilla-central/raw-file/$$SOURCEVERSION/security/manager/ssl/security-prefs.js \
+	https://hg.mozilla.org/mozilla-central/raw-file/$$SOURCEVERSION/modules/libpref/init/all.js \
+	https://hg.mozilla.org/mozilla-central/raw-file/$$SOURCEVERSION/testing/profiles/prefs_general.js \
+	https://hg.mozilla.org/mozilla-central/raw-file/$$SOURCEVERSION/layout/tools/reftest/reftest-preferences.js \
+	https://hg.mozilla.org/mozilla-central/raw-file/$$SOURCEVERSION/js/src/tests/user.js"; \
+	for SOURCEFILE in $$FIREFOX_SOURCE_PREFS; do wget "$$SOURCEFILE" -O - ; done | egrep "(^pref|^user_pref)" | sort --unique >| sourceprefs.js
+
+######################
+
+checknotcovered:
+	@# check for preferences present in firefox source but not covered by user.js
+	@# configure ignored preferences in ignore.list
+	@SOURCE_PREFS=$$(egrep '(^pref|^user_pref)' sourceprefs.js | awk -F'"' '{print $$2}'); \
+	for SOURCE_PREF in $$SOURCE_PREFS; do \
+	grep "\"$$SOURCE_PREF\"" user.js ignore.list >/dev/null || echo "Not covered by user.js : $$SOURCE_PREF"; \
+	done | sort --unique
+
+checkdeprecated:
+	@# check for preferences in hardened user.js that are no longer present in firefox source
+	@HARDENED_PREFS=$$(egrep "^user_pref" user.js | cut -d'"' -f2); \
+	for HARDENED_PREF in $$HARDENED_PREFS; do \
+	grep "\"$$HARDENED_PREF\"" sourceprefs.js  >/dev/null || echo "Deprecated : $$HARDENED_PREF"; \
+	done | sort --unique
+
+stats:
+	@# count preferences number, various stats
+	@echo "$$(egrep "^user_pref" user.js | wc -l | cut -f1) preferences in user.js"
+	@echo "$$(wc -l sourceprefs.js | cut -d" " -f1) preferences in Firefox source"
+
+cleanup:
+	@# remove temporary files
+	@# please comment this out when not needed, to minimize load on Mozilla servers
+	@rm sourceprefs.js
+
+authors:
+	@# generate an AUTHORS file, ordered by number of commits
+	@# TODO: add a .mailmap file to deduplicate authors with multiple email addresses
+	@# to add extra authors/credits, git commit --allow-empty --author="A U Thor <author@example.com>"
+	@git shortlog -sne | cut -f1 --complement >| AUTHORS
--- a/user.js
+++ b/user.js
@@ -368,6 +368,14 @@ user_pref("experiments.manifest.uri",				"");
 // PREF: Disallow Necko to do A/B testing
 user_pref("network.allow-experiments",			false);
 
+// PREF: Disable sending Firefox crash reports to Mozilla servers
+// https://wiki.mozilla.org/Breakpad
+// http://kb.mozillazine.org/Breakpad
+// https://dxr.mozilla.org/mozilla-central/source/toolkit/crashreporter
+// https://bugzilla.mozilla.org/show_bug.cgi?id=411490
+// A list of submitted crash reports can be found at about:crashes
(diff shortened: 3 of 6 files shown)


nodiscc:
Done with this part. Let me know if this should be split into separate PRs.

nodiscc:
Added complete coverage of `GEOLOCATION` and `QUIET FOX (1&2)` parts of #208 
See individual commits and messages

```
* 1a62106 disable Firefox Sync, fixes item 0380 of #208
* a770ea0 disable flyweb, fixes item 0376 of #208
* 70bbab1 Disable sending reports of tab crashes to Mozilla (about:tabcrashed), don't nag user about unsent crash reports: fixes items 0350 0351 of #208
* e776bf1 Disable sending Firefox crash reports to Mozilla servers, fixes item 0305 of #208
* a1bed5a Disallow Necko to do A/B testing, fixes item 0341 of #208, see https://dxr.mozilla.org/mozilla-central/source/modules/libpref/init/all.js#1350; set experiments manifest uri to blank, even though experiments are already disabled; update previously broken links in experiments/telemetry section fixes item 0340 of #208
* b62c59d disable flash player crash reports, don't send URLs in crash reports when enabled, fixes items 0309 0310 of #208
* 467fe9d opt-out of themes (persona) updates, fixes item 0307 of #208
* 3add13e Prevent leaking application locale/date format using JavaScript, fixes item 0208 of #208
* 76953bf add browser.search.geoSpecificDefaults* prefs, fixes item 0206 of #208
* 19c8579 add intl.locale.matchOS and general.useragent.locale, fixes items 0203 0204 of #208 more accurate description of accept_language setting```

**Comment by Thorin-Oakenpants** [comment]
@nodiscc : I see under [this](https://github.com/pyllyukko/user.js/pull/255/commits/c0962da8717aba4231e0fa7102e472b634d260f1) you listed a link for about: URLs

I find using `about:about` will display everything applicable to your profile.

Compared to the list at the URL, I am missing:
- about:app-manager - deprecated
- about:compartments - deprecated
- about:customizing - deprecated
- about:permissions - deprecated (although the page notes it was removed in FF45)
- about:sync-progress - valid entry, I have never set up sync
- about:reader - valid entry, I just have reader disabled

And here are six items (non-extension) not listed
- about:checkerboard
- about:debugging
- about:devtools-toolbox
- about:performance
- about:profiles
- about:serviceworkers

Might be better and easier to just use `about:about` in the readme, as it will always be current and correct regardless of version

**Comment by nodiscc** [comment]
Thanks, I edited https://developer.mozilla.org/en-US/Firefox/The_about_protocol and asked for review. It's better to link to that page since it's the actual documentation on all `about:` pages. `about:about:` is listed there. I added a note that available `about:` pages may vary depending on Firefox versions.

**Comment by nodiscc** [comment]
Results of `make tests` introduced in c37bf58 can be seen at https://gist.github.com/nodiscc/ed0c84d3492e7faf8869ed52b5647dbd

File: AUTHORS [hunk 1]
```diff
@@ -0,0 +1,16 @@
...
+Alex Hirsch <w4rh4wk@bluephoenix.at>
+DrunkenSasquatch <j.hamshere91@gmail.com>
+Frank LENORMAND <lenormf@gmail.com>
+Sebastian Schmidt <publicarray@icloud.com>
+Steffen Gransow <graste@mivesto.de>
+devmapper0 <devmapper0@users.noreply.github.com>
+Mehmet Atif Ergun <mehmetaergun@users.noreply.github.com>
+mengele-chan <mengele-chan@yandex.ru>
+uberspot <onexemailx@gmail.com>
+zummuz <zummuz@users.noreply.github.com>
```
pyllyukko: This is not needed. All the info is available as metadata.

user.js
```diff
@@ -325,16 +353,45 @@ user_pref("devtools.debugger.force-local",			true);
...
+// PREF: Disable FlyWeb (discovery of LAN/proximity IoT devices that expose a Web interface)
+// https://wiki.mozilla.org/FlyWeb
+// https://wiki.mozilla.org/FlyWeb/Security_scenarios
+// https://docs.google.com/document/d/1eqLb6cGjDL9XooSYEEo7mE-zKQ-o-AuDTcEyNhfBMBM/edit
+// http://www.ghacks.net/2016/07/26/firefox-flyweb
+user_pref("dom.flyweb.enabled",	false);
+
+// PREF: Disable Firefox Sync
+// https://wiki.mozilla.org/Services/Sync
+user_pref("services.sync.enabled",	false);
```
pyllyukko: I don't think this is necessary, as it's opt-in and some users might actually want to use it.

pyllyukko:
Will continue the review later, as I'm afraid I don't have the time to do it this week :frowning_face:

nodiscc:
@pyllyukko No problem, take your time, this can wait.

Note for the review: It will be *much* easier to add comments on individual commits (eg at https://github.com/pyllyukko/user.js/pull/255/commits/cdf0bf5fb381391dfca361609717997ac7115f67) instead of using the GH review review feature, where the scope is the entire PR. Commits are atomic, the whole PR is not.

Noted: 
 - [ ] remove authors file generation (note that the authorship information is lost when [downloading the ZIP file](https://github.com/pyllyukko/user.js#download), using doing a shallow clone `git clone --depth=...` or using `git-archive`)
 - [ ] Stop disabling Firefox sync (move to ignore.list)

user.js
```diff
@@ -170,13 +171,29 @@ user_pref("browser.search.countryCode",				"US");
...
 
+// PREF: Set Firefox locale to en-US
+user_pref("general.useragent.locale",			"en-US");
+
+// PREF: Don't use OS values to determine locale, force using Firefox locale setting
+user_pref("intl.locale.matchOS",				false);
+
+// PREF: Don't use Mozilla-provided location-specific search engines
+user_pref("browser.search.geoSpecificDefaults", 	false);
+user_pref("browser.search.geoSpecificDefaults.url",	"");
```
pyllyukko: Is the ```""``` necessary, when the feature is toggled off?

nodiscc: Probably not, this can be moved to ignore.list

Makefile
```diff
@@ -1,2 +1,56 @@
+all: whatdoesitdo tests authors
+
 whatdoesitdo:
+	@# generate the README "What does it do?" section
 	@./gen-readme.sh
+
+# To decrease tests verbosity, comment out unneeded targets
+tests: downloadffprefs checknotcovered checkdeprecated stats cleanup
```
pyllyukko: This will fail horribly if you have e.g. ```MAKEFLAGS="-j 4"```, like [I do](https://github.com/pyllyukko/dotfiles/blob/master/bashrc#L55).

I'm not entirely sure what is the proper way to fix it, as I'm not that familiar with makefiles. Maybe with [.NOTPARALLEL](https://www.gnu.org/software/make/manual/html_node/Special-Targets.html#Special-Targets)?

pyllyukko:
BTW. are you planning on pushing more commits into this? As I'm now [tweaking it](https://github.com/pyllyukko/user.js/tree/nodiscc-fixesfixesfixes) and trying to get it merged soon.

Makefile
nodiscc: Yes apparently a target named `.NOTPARALLEL` should inhibit the `-j` cli/config file option

pyllyukko: Cool, that's what I thought. I'm still not sure as to where and how to add it.

e.g., should it be like:

```
.NOTPARALLEL: downloadffprefs
```

Or just as it's own line? It wasn't completely clear to me.

nodiscc:
> are you planning on pushing more commits into this

No, please edit it as you like; I will add commit comments on https://github.com/pyllyukko/user.js/commits/nodiscc-fixesfixesfixes if needed

Makefile
nodiscc: I forgot to post the link I had found: http://stackoverflow.com/questions/4346399/how-can-i-force-gnu-make-to-not-build-recipe-in-parallel/23903293#23903293

```
.PHONY: all clean

.NOTPARALLEL:

anotherTarget: dependency1
```

Apparently the MAKEFLAGS variable can also be used (solution 3)

pyllyukko: Yeah I saw that post also when I was looking at the subject. Is it so, that everything below that will not run as parallel or what? I mean there are still some things, that can be run as parallel in the makefile, e.g. ```authors```, ```whatdoesitdo``` and all the rest for that matter, once ```downloadffprefs``` has finished.

nodiscc: > Is it so, that everything below that will not run as parallel or what?

This is what it looks like, but I have no experience with parallel usage of Make - very interesting, I really need to read the [full Make manual](https://www.gnu.org/software/make/manual/make.html).

Unfortunately no time to work on it at the moment, would you be ok to apply the global `MAKEFLAGS := --jobs=1` var, and open an issue (`Makefile: parallelize tests`) ? We can improve this later.

pyllyukko: Sounds good.

user.js
pyllyukko: 2a37932f655d8a90114294b4f7675cb27c0fb008

Makefile
```diff
@@ -1,2 +1,56 @@
...
+cleanup:
+	@# remove temporary files
+	@# please comment this out when not needed, to minimize load on Mozilla servers
+	@rm sourceprefs.js
+
+authors:
+	@# generate an AUTHORS file, ordered by number of commits
+	@# TODO: add a .mailmap file to deduplicate authors with multiple email addresses
+	@# to add extra authors/credits, git commit --allow-empty --author="A U Thor <author@example.com>"
+	@git shortlog -sne | cut -f1 --complement >| AUTHORS
```
pyllyukko: BTW. would the following ```.mailcap``` file be sufficient for this?

```
CHEF-KOCH <CHEF-KOCH@users.noreply.github.com> <Nvinside@gmail.com>
Francois Marier <francois@mozilla.com> <francois@debian.org>
```

nodiscc: s/mailcap/mailmap/

Yes it works with these contents (I just tested it with `git shortlog -sne`)

nodiscc: As far as I know there is no way to (directly) list users who posted in a github project issue tracker. But it can be parsed from https://api.github.com/repos/pyllyukko/user.js/issues (`login` values)

pyllyukko: It was a joke, we don't need to do that :)

PR #266, opened 2017-04-20
update README: remove unneeded links, clarify persistence

### LINKED ISSUE #247: Depending on installation method, settings may not be reset to default values when restarting Firefox
> However, the user.js file has this nice property, that even if you go change any of these settings through about:config, they're reset to the user.js defined values after you restart Firefox. So user.js makes sure they're back at the secure default values always when you start your browser.

This is only true for [Single profile installation](https://github.com/pyllyukko/user.js#single-profile-installation), this paragraph should be moved to that section/rewritten. 

In my case it is desirable that settings are not reset (and I know other users who would also like it this way). Is it possible to install user.js for a single profile, without locking the prefs? (Naming the file `prefs.js`? Untested)

nodiscc:
of settings across sessions depending on install method
fixes #247
wrap some lines to ~80 chars.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 **Firefox configuration hardening**
 
-A [user.js][1] configuration file for Mozilla Firefox designed to harden Firefox settings and
+A [user.js](http://kb.mozillazine.org/User.js_file) configuration file for [Mozilla Firefox](https://www.mozilla.org/en-US/firefox/new/) designed to harden browser settings and
 make it more secure.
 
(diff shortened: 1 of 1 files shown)


pyllyukko:
> wrap some lines to ~80 chars.

I don't think we need to wrap the README. It's a matter of the viewer/editor to do that.

nodiscc:
https://github.com/pyllyukko/user.js/pull/266/commits/7e4432a18c29563e4173c7ccfe8ead403b55f055

Issue #231, opened 2017-03-07
Question/discussion: relaxed settings variant/branch


pyllyukko:
So I was thinking of creating a bit more relaxed version of our user.js for other purposes / social media / "normal users". It would just be a separate branch in the repo (e.g. "relaxed") and with the current master being the default.

Here are some features that I thought could deviate from master:

* Commenting out ```browser.privatebrowsing.autostart```
  * Note that disk caching (and other similar settings) would still be disabled, so that no sensitive data would reside on disk
* Commenting out service workers & notifications settings
* Commenting out ```browser.urlbar.maxRichResults``` & ```browser.urlbar.autocomplete.enabled```
  * Adding ```browser.urlbar.suggest.history == false``` and making sure that only bookmarked sites and existing tabs are suggested if any
* Commenting out ```privacy.clearOnShutdown.cookies``` & altering ```network.cookie.lifetimePolicy```
  * Of course cookies are a major privacy issue, but users have the possibility to deal with this manually
* Commenting out ```privacy.clearOnShutdown.passwords``` so that Firefox's password manager can be used

So the actual questions here are:

* What do you think about having this kind of branch?
* What other settings/features "normal users" actually use? :smiley:

Atavic:
Do they normally start with a private browsin?
```
user_pref("browser.privatebrowsing.autostart", true);
```
Do they normally delete their cookies?
```
user_pref("privacy.clearOnShutdown.cookies", true);
```

Roman-Nopantski:
Gee @pyllyukko , we discussed just that the other day on irc. I think it's a great idea because I said I was planning on doing it - I am literally weeks away from providing three branch templates from `tiny-if-any-breakage-must-have-settings` to `current` to `the-bastard-uber-hardened`. Way to to go bro.

nodiscc:
@pyllyukko Once #229 is solved, I intend to propose some changes from my adapted user.js, these changes mostly fall in the following categories:

 * Re-enable caching/password storage/histories, don't use private browsing by default (improves performance, reduces amount of data transfered over the network, improves usability. Full Disk Encryption mitigates some threats related to storing this kind of personal information.)
 * Re-enable usability features that don't leak data over the network (re-enable autocompletion and related. Basically remove mitigation of the shoulder-surfing threat.)

One question remains, where/how to store/distribute the relaxed settings:

 * As a separate file `relaxed.js` containing all the changes?
[...]

PR #267, opened 2017-04-24
add various preferences to user.js/ignore.list:

nodiscc:
user.js:
 * fallback on mozilla wifi geolocation services when user re-enables geolocation (instead of Google services)
 * disable geolocation requests logging in console
 * disable more Virtual Reality APIs
 * disable vibrator API
 * when user re-enables URL fixup, still disable it for URLs using 'username:password' syntax
 * enable downloading/using Mozilla SWF blocklists in case some users actually need to re-enable Flash player plugin
 * use master switch browser.selfsupport.enabled to disable Heartbeat, in addition to already empty URL.
 * improve safebrowsing.downloads.enabled description
 * add "Open windows" to default selection in "Clear Private Data" dialog
 * enforce disabling fully automatic password autofill on HTTP sites, even when user re-enables autofill
 * enforce showing an in-content warning for insecure login forms

ignore.list:
 * add browser.safebrowsing.downloads.* fine controls
 * add safebrowsing url prefs
 * explicitely don't set/harden privacy.cpd.passwords/siteSettings and mark it: '(relaxed) ($REASON)' - in this case, reason is better usability - Ref. #231

Thanks to https://github.com/ghacksuserjs/ghacks-user.js/

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,8 @@ HTML5 / [APIs](https://wiki.mozilla.org/WebAPI) / [DOM](https://en.wikipedia.org
 * Make sure the User Timing API does not provide a new high resolution timestamp [ [1](https://trac.torproject.org/projects/tor/ticket/16336) ]
 * Disable Web Audio API [ [1](https://bugzilla.mozilla.org/show_bug.cgi?id=1288359) ]
 * Disable Location-Aware Browsing (geolocation) [ [1](https://www.mozilla.org/en-US/firefox/geolocation/) ]
+* When geolocation is enabled, use Mozilla geolocation service instead of Google [ [1](https://bugzilla.mozilla.org/show_bug.cgi?id=689252) ]
+* When geolocation is enabled, don't log geolocation requests to the console
 * Disable raw TCP socket support (mozTCPSocket) [ [1](https://trac.torproject.org/projects/tor/ticket/18863) [2](https://www.mozilla.org/en-US/security/advisories/mfsa2015-97/) [3](https://developer.mozilla.org/docs/Mozilla/B2G_OS/API/TCPSocket) ]
 * Whether JS can get information about the network/browser connection
 * Disable WebRTC entirely to prevent leaking internal IP addresses (Firefox < 42)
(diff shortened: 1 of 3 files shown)


nodiscc:
* doc: add reference link for geo.wifi.*:
these preferences are no longer set by default but are still readable and will work correctly:
> The preference is still honored, it is just not in about:config.
> In about:config it should still be possible to create a string preference with that name, and it should work correctly.

 * add WONTFIX items from #208 to ignore.list, up to HARDWARE FINGERPRINTING section;
 * move extra VR API preferencs to ignore.list since VR is globally disabled
 * move New Tab Page/Tiles prefs next to each other

nodiscc:
* 341f024 ignore network.http.redirection-limit, disable in-content SVG rendering item 2671 of #208 states this pref should not be altered for convenience purposes: disable SVG spport anyway and add a clear notice about breaking functionality this setting should be removed in "relaxed" variants; ref #231
* 4ee5f90 add tab warnings to ignore.list
* e419f21 add more ignored settings, add browser.shell.shortcutFavicons remove misplaced pref in ignore.list thanks https://github.com/ghacksuserjs/ghacks-user.js
* b15c546 add more ignored prefs (UI related) thanks https://github.com/ghacksuserjs/ghacks-user.js move DOM Timing API options next to each other
* 372f18e Disable Web Workers TODO: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers#Other_types_of_worker investigate disabling chromeworkers/audioworkers
* a62b05a add more prefs to ignore list (disk cache)
* 80309fc enforce force Firefox update checks, clarify auto update notice, remove deprecated note (can be found through make checkdeprecated)

nodiscc:
@pyllyukko I'm done with this PR. Let me know if anything should be changed. I will submit future changes to a separate Pull Request.

nodiscc:
This PR is confusing and covers several unrelated aspects.
I will close it now and reopen separate Pull Requests for each individual change.
Note that I will not run `make whatdoesitdo` to ease merging/reviewing. @pyllyukko You'll have to run it after merging.

pyllyukko:
> This PR is confusing and covers several unrelated aspects.
> I will close it now and reopen separate Pull Requests for each individual change.

git_horror

Reminds me of this 🤣 

guitar_hero2

PR #276, opened 2017-04-28
Disable in-content SVG rendering

nodiscc:
Can be considered for relaxing #231

### DIFF
--- a/user.js
+++ b/user.js
@@ -254,6 +254,13 @@ user_pref("javascript.options.asmjs",				false);
 // https://github.com/iSECPartners/publications/tree/master/reports/Tor%20Browser%20Bundle
 user_pref("gfx.font_rendering.opentype_svg.enabled",		false);
 
+// PREF: Disable in-content SVG rendering (Firefox >= 53)
+// NOTICE: Disabling SVG support breaks many UI elements on many sites
+// https://bugzilla.mozilla.org/show_bug.cgi?id=1216893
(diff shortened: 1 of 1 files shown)


TriMoon:
Why would one want to do this??
The only reason i can think of is embedded javascript inside the SVG, but that can be blocked by other means eg Content-Policy and blocking addons like uMatrix.

pyllyukko:
There has been vulnerabilities in SVG parsing, so I wouldn't consider it as harmless. It's about reducing attack surface.

I do realize, that this might break bunch of stuff as was also noted [here](https://github.com/pyllyukko/user.js/issues/287#issuecomment-298479771). And we might need to revisit this setting.

If you check [Firefox security advisories](https://www.mozilla.org/en-US/security/known-vulnerabilities/firefox/), there's a whole bunch of SVG vulns. [#258511](https://bugzilla.mozilla.org/show_bug.cgi?id=258511) tells the same story and that's why the preference exists.

TriMoon:
Thanks i wasn't aware of that (fixed) exploit in svg... 👍

nodiscc:
For reference there was at least one (fixed) severe security issue with SVG in Firefox: https://www.mozilla.org/en-US/security/advisories/mfsa2016-92/ (Firefox SVG Animation Remote Code Execution - Firefox 50.0.2, ESR 45.5.1)

PR #279, opened 2017-04-28
harden bookmarks functionality:

nodiscc:
* Don't fetch and permanently store bookmarks favicons to $profile_dir/shortcutCache
 * Disable periodic bookmarks backups (default: 15) to disk

### DIFF
--- a/user.js
+++ b/user.js
@@ -717,8 +717,9 @@ user_pref("browser.helperApps.deleteTempFileOnExit",		true);
 // https://developer.mozilla.org/en-US/docs/Mozilla/Preferences/Preference_reference/browser.pagethumbnails.capturing_disabled
 user_pref("browser.pagethumbnails.capturing_disabled",		true);
 
-// PREF: Don't fetch and permanently store bookmarks favicons to $profile_dir/shortcutCache
-// NOTICE: Bookmarks will only be shown with a generic icon
+// PREF: Don't fetch and permanently store favicons for Windows .URL shortcuts created by drag and drop
+// NOTICE: .URL shortcut files will be created with a generic icon
+// Favicons are stored as .ico files in $profile_dir\shortcutCache
 user_pref("browser.shell.shortcutFavicons",					false);
 
 // PREF: Disable bookmarks backups (default: 15)
--- a/user.js
+++ b/user.js
@@ -717,6 +717,14 @@ user_pref("browser.helperApps.deleteTempFileOnExit",		true);
 // https://developer.mozilla.org/en-US/docs/Mozilla/Preferences/Preference_reference/browser.pagethumbnails.capturing_disabled
 user_pref("browser.pagethumbnails.capturing_disabled",		true);
 
+// PREF: Don't fetch and permanently store bookmarks favicons to $profile_dir/shortcutCache
+// NOTICE: Bookmarks will only be shown with a generic icon
+user_pref("browser.shell.shortcutFavicons",					false);
+
+// PREF: Disable bookmarks backups (default: 15)
+// http://kb.mozillazine.org/Browser.bookmarks.max_backups
+user_pref("browser.bookmarks.max_backups", 0);
+
 /*******************************************************************************
  * SECTION: UI related                                                         *
  *******************************************************************************/


user.js
```diff
@@ -717,6 +717,14 @@ user_pref("browser.helperApps.deleteTempFileOnExit",		true);
 // https://developer.mozilla.org/en-US/docs/Mozilla/Preferences/Preference_reference/browser.pagethumbnails.capturing_disabled
 user_pref("browser.pagethumbnails.capturing_disabled",		true);
 
+// PREF: Don't fetch and permanently store bookmarks favicons to $profile_dir/shortcutCache
+// NOTICE: Bookmarks will only be shown with a generic icon
+user_pref("browser.shell.shortcutFavicons",					false);
```
pyllyukko: Is this necessary?

Thorin-Oakenpants:
It's not bookmark shortcuts. It's shortcuts icons cached for the system to use with URL shortcuts - eg drag and drop an url to your dekstop.

Bookmark icons are stored in the sql tables

```js
/** FAVICONS ***/
/* 1030: disable favicons in shortcuts
 * URL shortcuts use a cached randomly named .ico file which is stored in your
 * profile/shortcutCache directory. The .ico remains after the shortcut is deleted.
 * If set to false then the shortcuts use a generic Firefox icon ***/
user_pref("browser.shell.shortcutFavicons", false);
/* 1031: disable favicons in tabs and new bookmarks
 * bookmark favicons are stored as data blobs in places.sqlite>moz_favicons ***/
   // user_pref("browser.chrome.site_icons", false);
   // user_pref("browser.chrome.favicons", false);
/* 1032: disable favicons in web notifications ***/
   // user_pref("alerts.showFavicons", false);
```

nodiscc:
@Thorin-Oakenpants Thanks for the hint, my understanding of this pref was wrong. I will amend the pref description

Is this behavior Windows-specific? When I drag an URL/bookmark to my desktop/file manager (XFCE4/xfdesktop+thunar on Debian) it simply creates a standard [Freedesktop .desktop launcher](https://standards.freedesktop.org/desktop-entry-spec/latest/) using a generic icon `Icon=gnome-fs-bookmark`. No `$profile_dir/shortcutCache` is present.

Then the description should be

```
// PREF: Don't fetch and permanently store favicons for Windows .URL shortcuts created by drag and drop
// NOTICE: .URL shortcut files will be created with a generic icon
// Favicons are stored as .ico files in $profile_dir/shortcutCache
```

@pyllyukko I believe we can safely ignore `browser.chrome.site_icons, browser.chrome.favicons` since they don't cause useless persistent caching?

Thorin-Oakenpants:
The other inactive prefs in the code sample above we left in for informational purposes (some people don't like favicons). The alerts I'm not sure on. There's no real damage having randomly named icons cached (there's no info in them), but it breaks that persistent storage and crosses the browser/os line. It also never cleans itself AFAIK. I had several hundred wiki (and some other site icons) in mine covering about 2 years (I always add a wikipedia link to the root folders of tv shows in my media drives). I guess if someone ever got onto my pc they could infer I had been to wikipedia based on the icon image - big deal, they're IN my PC :)

Windows only? Not 100% sure. We're going to go thru all diff between Win/Mac/Nix and do some tagging - see https://github.com/ghacksuserjs/ghacks-user.js/issues/109 - see erthling's dump here-> https://earthlng.github.io/FFprefs-diffs/Firefox-v53.0.html . Note: `browser.shell.shortcutFavicons` is not listed in that diff, so at least the pref exists in all flavors - just need to clarify the behavior.

We also discussed and tested in depth favicons - see https://github.com/ghacksuserjs/ghacks-user.js/issues/48 and also mentioned/covered/more-tests in https://github.com/ghacksuserjs/ghacks-user.js/issues/48

OSX - our MAC man is @grauenwolfe who is a collaborator at ghacks - just throw him a comment

Thorin-Oakenpants:
Oh, one thing I didn't think of, and I'm not sure - but fav icons can be specified - eg cross domain. This is a definite possible tracking mechanism, and of the top of my head I am not sure if uMatrix, uBo covers these?

Atavic:
uMatrix sees them as images:
`03:08:18 image https://s2.googleusercontent.com/s2/favicons`

user.js
Atavic: There's some PoC where a phishing google shortcut is put on the desktop.

nodiscc:
@pyllyukko Should it be merged? Or should I move these 2 prefs to ignore.list? Thanks

PR #281, opened 2017-04-28
When geolocation is enabled, use Mozilla geolocation service instead of Google

nodiscc:


user.js
```diff
@@ -32,6 +32,13 @@ user_pref("dom.webaudio.enabled",				false);
 // https://www.mozilla.org/en-US/firefox/geolocation/
 user_pref("geo.enabled",					false);
 
+// PREF: When geolocation is enabled, use Mozilla geolocation service instead of Google
+// https://bugzilla.mozilla.org/show_bug.cgi?id=689252
+user_pref("geo.wifi.uri", "https://location.services.mozilla.com/v1/geolocate?key=test");
```
fmarier: Isn't this the default?

At least it is on Firefox 52 on my Debian machine and in Firefox Nightly.

Thorin-Oakenpants: On my Windows & FF53 the default is "https://www.googleapis.com/geolocation/v1/geolocate?key=%GOOGLE_API_KEY%"

fmarier: Interesting, as far as I can see:

- release and beta default to Google
- nightly defaults to Mozilla
- Debian defaults to Mozilla regardless

fmarier: Instead of `key=test`, we may as well use the built-in Mozilla Location Service key:

    https://location.services.mozilla.com/v1/geolocate?key=%MOZILLA_API_KEY%

It's more likely to work reliably.

Thorin-Oakenpants: I had a look in FF55, same story. Yeah, I noticed the mozilla api key, it's used in

`browser.search.geoip.url`
https://location.services.mozilla.com/v1/country?key=%MOZILLA_API_KEY%

PR #285, opened 2017-04-28
Prevent sites from using external links like mailto.

SvobodaJakub:
- Disable protocols like mailto from being handled by external applications, so that evil trap sites can't launch external applications.

### DIFF
--- a/user.js
+++ b/user.js
@@ -270,7 +270,7 @@ user_pref("general.buildID.override",				"20100101");
 // https://github.com/pyllyukko/user.js/issues/120
 user_pref("browser.display.use_document_fonts",			0);
 
-// PREF: Prevent sites from using links like mailto to launch external applications.
+// PREF: Prevent sites from using URLs such as mailto:, irc:, magnet:, ... to launch external applications
 // http://kb.mozillazine.org/Network.protocol-handler.external-default
 // http://kb.mozillazine.org/Network.protocol-handler.warn-external-default
 // https://news.ycombinator.com/item?id=13047883
--- a/user.js
+++ b/user.js
@@ -304,7 +304,7 @@ user_pref("network.protocol-handler.warn-external.vnd.youtube", true);
 // http://kb.mozillazine.org/Network.protocol-handler.expose.%28protocol%29
 // NOTICE: Disabling nonessential protocols breaks all interaction with custom protocols such as mailto: irc: magnet: ...
 // TODO: Add more protocols, see TODO for the previous PREF.
-// network.protocol-handler.expose-all == false breaks some javascript links, hence not used.
+// network.protocol-handler.expose-all == false breaks ordinary a href links, hence not used.
 user_pref("network.protocol-handler.expose.irc", 	false);
 user_pref("network.protocol-handler.expose.magnet", 	false);
 user_pref("network.protocol-handler.expose.mailto", 	false);
--- a/user.js
+++ b/user.js
@@ -270,7 +270,7 @@ user_pref("general.buildID.override",				"20100101");
 // https://github.com/pyllyukko/user.js/issues/120
 user_pref("browser.display.use_document_fonts",			0);
 
-// PREF: Prevent sites from using external links like mailto.
+// PREF: Prevent sites from using links like mailto to launch external applications.
 // http://kb.mozillazine.org/Network.protocol-handler.external-default
 // http://kb.mozillazine.org/Network.protocol-handler.warn-external-default
 // https://news.ycombinator.com/item?id=13047883
@@ -279,20 +279,43 @@ user_pref("browser.display.use_document_fonts",			0);
 // TODO: Add externally-handled protocols from Windows 8.1 and Windows 10 (currently contains protocols only from Linux and Windows 7) that might pose a similar threat (see e.g. https://news.ycombinator.com/item?id=13044991)
 // TODO: Add externally-handled protocols from Mac OS X that might pose a similar threat (see e.g. https://news.ycombinator.com/item?id=13044991)
 user_pref("network.protocol-handler.external-default", 		false);
+user_pref("network.protocol-handler.external.irc", 		false);
+user_pref("network.protocol-handler.external.magnet", 		false);
 user_pref("network.protocol-handler.external.mailto", 		false);
 user_pref("network.protocol-handler.external.ms-windows-store", false);
 user_pref("network.protocol-handler.external.news", 		false);
 user_pref("network.protocol-handler.external.nntp", 		false);
+user_pref("network.protocol-handler.external.sftp", 		false);
 user_pref("network.protocol-handler.external.snews", 		false);
 user_pref("network.protocol-handler.warn-external-default", 	true);
+user_pref("network.protocol-handler.warn-external.irc", 	true);
+user_pref("network.protocol-handler.warn-external.magnet", 	true);
 user_pref("network.protocol-handler.warn-external.mailto", 	true);
 user_pref("network.protocol-handler.warn-external.ms-windows-store", true);
 user_pref("network.protocol-handler.warn-external.news", 	true);
 user_pref("network.protocol-handler.warn-external.nntp", 	true);
+user_pref("network.protocol-handler.warn-external.sftp", 	true);
 user_pref("network.protocol-handler.warn-external.sms", 	true);
 user_pref("network.protocol-handler.warn-external.snews", 	true);
 user_pref("network.protocol-handler.warn-external.tel", 	true);
 user_pref("network.protocol-handler.warn-external.vnd.youtube", true);
+
+// PREF: Disable nonessential protocols from being used altogether.
+// http://kb.mozillazine.org/Network.protocol-handler.expose.%28protocol%29
+// NOTICE: Breaks all interactions and links using these protocols.
+// TODO: Add more protocols, see TODO for the previous PREF.
+// network.protocol-handler.expose-all == false breaks some javascript links, hence not used.
+user_pref("network.protocol-handler.expose.irc", 	false);
+user_pref("network.protocol-handler.expose.magnet", 	false);
+user_pref("network.protocol-handler.expose.mailto", 	false);
+user_pref("network.protocol-handler.expose.ms-windows-store", false);
+user_pref("network.protocol-handler.expose.news", 	false);
+user_pref("network.protocol-handler.expose.nntp", 	false);
+user_pref("network.protocol-handler.expose.sftp", 	false);
+user_pref("network.protocol-handler.expose.sms", 	false);
+user_pref("network.protocol-handler.expose.snews", 	false);
+user_pref("network.protocol-handler.expose.tel", 	false);
+user_pref("network.protocol-handler.expose.vnd.youtube", false);
 
 /******************************************************************************
  * SECTION: Extensions / plugins                                                       *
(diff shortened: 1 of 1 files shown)


nodiscc:
@SvobodaJakub there are also `irc` and `magnet` protocol handlers

user.js
```diff
@@ -270,6 +270,53 @@ user_pref("general.buildID.override",				"20100101");
 // https://github.com/pyllyukko/user.js/issues/120
 user_pref("browser.display.use_document_fonts",			0);
 
+// PREF: Prevent sites from using links like mailto to launch external applications.
+// http://kb.mozillazine.org/Network.protocol-handler.external-default
+// http://kb.mozillazine.org/Network.protocol-handler.warn-external-default
+// https://news.ycombinator.com/item?id=13047883
+// https://bugzilla.mozilla.org/show_bug.cgi?id=167475
+// NOTICE: Breaks mailto links on legit websites, user has to right-click and copy mail address from the link.
```
nodiscc: Currently `NOTICE:` needs to be on the line just after `PREF:` for the automatic doc generation to work properly :/

user.js
```diff
@@ -270,6 +270,53 @@ user_pref("general.buildID.override",				"20100101");
...
+user_pref("network.protocol-handler.warn-external.nntp", 	true);
+user_pref("network.protocol-handler.warn-external.sftp", 	true);
+user_pref("network.protocol-handler.warn-external.sms", 	true);
+user_pref("network.protocol-handler.warn-external.snews", 	true);
+user_pref("network.protocol-handler.warn-external.tel", 	true);
+user_pref("network.protocol-handler.warn-external.vnd.youtube", true);
+
+// PREF: Disable nonessential protocols from being used altogether.
+// http://kb.mozillazine.org/Network.protocol-handler.expose.%28protocol%29
+// NOTICE: Breaks all interactions and links using these protocols.
```
nodiscc: Same as above

user.js
```diff
@@ -270,6 +270,53 @@ user_pref("general.buildID.override",				"20100101");
 // https://github.com/pyllyukko/user.js/issues/120
 user_pref("browser.display.use_document_fonts",			0);
 
+// PREF: Prevent sites from using links like mailto to launch external applications.
```
nodiscc: `// PREF: Prevent sites from using URLs such as mailto:, irc:, magnet: ... to launch external applications`

user.js
nodiscc: Should be `// NOTICE: Breaks opening third-party mail/messaging/torrent/... clients by clicking on links with custom protocols`

SvobodaJakub:
@nodiscc Thanks, you've inspired me to add a second block of .\*expose.\*==false which I'd like to discuss. The documentation seems to suggest that expose.(protocol)==false kills the protocol altogether. When testing with a mailto: link, that doesn't work and a popup opens under these conditions:

FF 53.0
user_pref("network.protocol-handler.expose-all", true); // default
user_pref("network.protocol-handler.expose.mailto", false); // default
user_pref("network.protocol-handler.external-default", false);
user_pref("network.protocol-handler.external.mailto", true); // default
user_pref("network.protocol-handler.warn-external-default", true);
user_pref("network.protocol-handler.warn-external.mailto", false); // default

Some other combinations of the prefs behave the same and the common thing is that either value of network.protocol-handler.expose.mailto prevents nothing.

Since user_pref("network.protocol-handler.expose-all", false); breaks javascript links, it seems that all the .\*expose.\* options are useless and I should remove them. Is that correct?

user.js
nodiscc: Should be `// NOTICE: Breaks all interaction with custom protocols such as mailto: irc: magnet: ...`

nodiscc:
> it seems that all the .*expose.* options are useless and I should remove them. Is that correct?

Do I understand correctly, does `"network.protocol-handler.expose-all", false` disable all custom protocol URLs, including `javascript:` links?

user.js
SvobodaJakub: I generated the readme using gen-readme.sh and the NOTICEs, the PREFs and their URLs were all added correctly to README.md. Am I doing something wrong?

nodiscc: Sorry. You are correct. Reference links must go right under the `PREF`. Everything is fine.

SvobodaJakub:
> Do I understand correctly, does "network.protocol-handler.expose-all", false disable all custom protocol URLs, including javascript: links?

I just tested it again (because the last time I did it was some time ago and I apparently made a mistake) and at least some javascript: links work but all a href links stopped working for me.

user.js
SvobodaJakub: Thanks for the wording suggestions.

nodiscc:
Ideally what we want is:

 * prevent opening any third party application custom protocols in URLs
 * ideally show a warning dialog (is that what `*.warn-external.*` is for?) when blocking such links
 * do not break basic, internal browser functionality (http/https/javascript: links)
 * have the ability to easily whitelist some useful protocols (commented out by default)

Feel free to experiment and update your patch + pref descriptions, I will have a look soon.

SvobodaJakub:
I made experiments and googled and it's a mess.

**network.protocol-handler.warn-external-default**

For protocol handlers that are set as external (described in the next section) and for which no default action has been set by the user, and a protocol-specific boolean `network.protocol-handler.warn-external.(protocol)` doesn't exist, it determines whether to ask before opening such a link or whether to automatically open the link in an application that can handle the protocol.

Example:

If http protocol is not exposed (described in a next section, basically disabled from being handled internally as usual), and FF settings are otherwise default, then if `network.protocol-handler.warn-external-default;false`, http links are opened in a new tab (it calls firefox, firefox communicates back with the first instance and tells it to open this link and terminates, the original firefox instance opens the link in a new tab); if `network.protocol-handler.warn-external-default;true`, then a window asking how to open appears.

The example can be tried by setting default clean FF to `network.protocol-handler.expose-all;false` and trying out http/https links with `network.protocol-handler.warn-external-default;{true,false}` and seeing the difference.

This option collides with a "newer" functionality for determining default actions so that once the user selected a default way to handle the protocol, no warning is ever issued again - https://bugzilla.mozilla.org/show_bug.cgi?id=440892

**network.protocol-handler.external-default**

For protocols for which a protocol-specific boolean `network.protocol-handler.external.(protocol)` doesn't exist, and they are not handled internally already, it determines whether the protocol can be handled at all.

Example: 

If http protocol is not exposed (described in a next section, basically disabled from being handled internally as usual), and FF settings are otherwise default, then if `network.protocol-handler.external-default;true`, http links produce a window that allows them to be opened in firefox (or other apps if other http-handling apps are on the system); if `network.protocol-handler.external-default;false`, then clicking a http link does nothing.

The example can be tried by setting default clean FF to `network.protocol-handler.expose-all;false` and trying out http/https links with `network.protocol-handler.external-default;{true,false}` and seeing the difference. Then set `network.protocol-handler.external-default` to the default value (`network.protocol-handler.external-default;true`) and create a new boolean pref `network.protocol-handler.external.http;false`; see that http links stopped working at all but https links produce the open-with window. Then see that the http/https behavior switches around when you switch around these two prefs (`network.protocol-handler.external-default;false` and `network.protocol-handler.external.http;true`, http links produce a window, https links do nothing).

The documentation at http://kb.mozillazine.org/Network.protocol-handler.external-default seems misleading or very unclear to me.

**network.protocol-handler.expose-all**

For protocols for which a protocol-specific boolean `network.protocol-handler.expose.(protocol)` doesn't exist, and they are not handled externally already (see previous sections), it determines whether the protocol can be handled at all.

Example:

If `network.protocol-handler.expose-all;false` and otherwise FF is in default settings, http links do not work (they are not handled externally according to the settings) but mailto links do work (they are handled externally). Adding a new boolean value `network.protocol-handler.external.http;true` makes http links to start sort-of working (see the previous section). Setting `network.protocol-handler.external.mailto;false` makes mailto links to stop working.

**Conclusion**

`.*expose.*`, `.*external.*`, and `.*warn-external.*` interact in complicated ways and they also interact with preferences from about:preferences#applications. A shotgun approach of setting expose.(proto)=false, external.(proto)=false, warn-external.(proto)=true seems like the best bet to disable a particular protocol.

SvobodaJakub:
A matter of further research might be setting `network.protocol-handler.expose-all;false` and setting common protocols (http, https, javascript) to `network.protocol-handler.expose.(protocol);true`, thus whitelisting protocols instead of the current blacklisting. There might be dragons ahead so I'm not putting that to the pull request, I'm going to dogfood this whitelist for a while.

pyllyukko:
Maybe this could be of assistance: [
Discover and Hack URL handlers](https://github.com/ouspg/urlhandlers)?

SvobodaJakub:
Thanks for the pointer to the urlhandlers project.

To clarify:

- Blacklist of known possibly useless protocols - The current changes in this pull request work and I am using them (and have been using them in a very similar form for a year already) - this is what I'd like to contribute.

- Whitelist (mentioned in https://github.com/pyllyukko/user.js/pull/285#issuecomment-298124944) - I have encountered problems when Firefox decided to delete the added *expose.(protocol) flags, breaking everything until I toggled `network.protocol-handler.expose-all` back to default. Definitely needs more testing to find out which methods of applying user.js are stable and which are not. - I don't want to contribute this at the moment and this is not in this pull request.

- Additional protocols as mentioned in https://github.com/pyllyukko/user.js/pull/285#issuecomment-299272781 - I can either do research into existing lists of all known protocols to create hundreds of prefs blacklisting them and add them to this pull request or we can get this merged as it is and I can submit another PR with the huge blacklist later. Either way, it'll take me a few weeks probably.

pyllyukko:
So how complete is the blacklist? Of course the whitelisting would be better and we should try to get that working in the future.

pyllyukko:
My first experiments with whitelisting:

* Set ```network.protocol-handler.expose-all``` to ```false``` -> everything breaks
* Set ```network.protocol-handler.external.http(s)``` to ```false``` -> Firefox no longer asks what app to use for https etc., but things still don't work
* Set ```network.protocol-handler.expose.http(s)``` to ```true``` -> things start to work

Need to experiment more. Javascript seemed to work all the time.

pyllyukko:
> I have encountered problems when Firefox decided to delete the added *expose.(protocol) flags, breaking everything until I toggled network.protocol-handler.expose-all back to default.

Sounds like there has been similar issues elsewhere also: [Why does Firefox set network.protocol-handler.expose.(protocol) to "true" on every start?](https://support.mozilla.org/en-US/questions/933326)

pyllyukko:
This is a very welcome enhancement for the user.js, but I would like to see it as whitelisting. I think I'll start eating some dog food too with <https://github.com/pyllyukko/user.js/tree/protocol-handler-whitelisting> and see what happens.

pyllyukko:
I think I'll merge <https://github.com/pyllyukko/user.js/tree/protocol-handler-whitelisting> after a bit more testing, if I don't encounter any major issues.

Could you guys test this as well?

pyllyukko:
FYI: [BeEF](http://beefproject.com/) uses protocol handlers to [fingerprint](https://github.com/beefproject/beef/wiki/Module:-browser-fingerprint) Firefox. My browser gets detected as "1+,10+" with ```moz-icon://.autoreg?size=16``` & ```chrome://browser/skin/sync-128.png```. Setting ```network.protocol-handler.expose.(protocol)``` to ```false``` doesn't seem to help 🙁

pyllyukko:
This is weird (from ```make downloadffprefs```):

```
pref("network.protocol-handler.external.ile", false);  // file
pref("network.protocol-handler.external.le", false);   // file
pref("network.protocol-handler.external.ps", false);   // https
pref("network.protocol-handler.external.tps", false);  // https
pref("network.protocol-handler.external.ttp", false);  // http
pref("network.protocol-handler.external.ttps", false); // https
```

pyllyukko:
I merged the [protocol-handler-whitelisting](https://github.com/pyllyukko/user.js/commits/protocol-handler-whitelisting) branch.

PR #300, opened 2017-05-21
[master] add NOTICE for browser.display.use_document_fonts

nodiscc:
Aside from the minor inconvenience of having all pages display with default Serif/Sans Serif fonts (you get used to it), `browser.display.use_document_fonts = false` breaks functionality on websites using icon fonts (such as [font awesome](http://fontawesome.io/)) to display UI elements/buttons/symbols/...

This is one of the main complaints I've heard from `user.js` users IRL/online. I have disabled this myself. A a _lot_ of websites make heavy use of icon fonts, sometimes there's an `alt` text or other indication that this or that is the button you are looking for, but most often not. As it's only used to prevent fingerprinting I suggest that it can be disabled in https://github.com/pyllyukko/user.js/tree/relaxed. Should I send another patch?

### DIFF
--- a/user.js
+++ b/user.js
@@ -306,7 +306,8 @@ user_pref("media.video_stats.enabled",				false);
 // https://bugzilla.mozilla.org/show_bug.cgi?id=583181
 user_pref("general.buildID.override",				"20100101");
 
-// PREF: Prevent font fingerprinting
+// PREF: Don't let websites specify which fonts to use to prevent font fingerprinting
+// NOTICE: Disabling document fonts breaks functionality websites using custom icon fonts
 // https://browserleaks.com/fonts
 // https://github.com/pyllyukko/user.js/issues/120
 user_pref("browser.display.use_document_fonts",			0);


Thorin-Oakenpants:
> breaks functionality on websites using icon fonts

AFAIK, this does not block icon fonts / glyphs : which is `gfx.downloadable_fonts.enabled`. I have `document_fonts` set to 0 - thus limiting myself to the basic 3 sets. And yet I can still get awesome font. I actually block all these as well (via uBo using `no-remote-fonts: * true` and then I can whitelist the few sites where it's needed (most sites for me I can easily guess the function of the missing button/icon/font).

But as for a relaxed branch, for sure (up to you guys). I just don't agree with your argument as to why.

nodiscc:
Strange, with `gfx.downloadable_fonts.enabled = true` but `browser.display.use_document_fonts = false` I still get broken icons on many sites (eg https://imgur.com/)

Thorin-Oakenpants:
You must be blocking it via some other mechanism - maybe something in NS?

http://terrillthompson.com/tests/icomoon/ - this is a simple test page

I have
- fonts:`browser.display.use_document_fonts = false`
- glyphs: `gfx.downloadable_fonts.enabled = true`
- uBo blocks all remote fonts: `no-remote-fonts: * true`

I see broken icons for the player controls. But if I add an exception to uBo: `no-remote-fonts: terrillthompson.com false` - and refresh, I can see the play, forward and stop buttons.

nodiscc:
It is unclear what's going on: 

 * I can see the glyphs on the test page http://terrillthompson.com/tests/icomoon/ *even* with `browser.display.use_document_fonts + gfx.downloadable_fonts.enabled = false`, and uBlock font blocking ON.
 * With `browser.display.use_document_fonts = 0` some icons from the uBlock popup menu are not displayed correctly (lock and refresh icons, a dot character is displayed instead), and glyphs are broken on many sites (for example the `?` icon top-right of [this page](https://dxr.mozilla.org/mozilla-esr45/)). Setting `browser.display.use_document_fonts = 1` fixes these problems.
 * I can never see the top button glyphs on https://imgur.com/ regardless of my settings.

Do you have the same results?

Thorin-Oakenpants:
> Do you have the same results?

Maybe we need to use a nilla firefox and work out what is being set/accessed how. Is it css font-face etc? Not a font expert, but there are a couple of ways to get fonts, right?. Maybe imgur does this differently? I will come back to this, and test imgur later - bookmarked this page so I don't forget. Or open an issue on ghacks repo, rather than in here, so I can document steps, add screenshots. Wonder how FF decides what is a glyph? What font is imgur using for icons? 9note, will come back to this topic, just can't do it right now)

> With browser.display.use_document_fonts = 0 some icons from the uBlock popup menu are not displayed correctly

Mine are. Wonder what difference uBo does for Linux vs Windows. Is there anything in uBo or uMatrix forums?

pyllyukko:
> http://terrillthompson.com/tests/icomoon/ - this is a simple test page

I have:

* ```browser.display.use_document_fonts = 0```
* ```gfx.downloadable_fonts.enabled = true```
* uBlock "Block remote fonts" -> false

I get:

iconfont

Imgur:

imgur

Imgur (with ```browser.display.use_document_fonts = 1```:

imgur2

nodiscc:
@pyllyukko your results are the same as mine, so it's safe to assume that `browser.display.use_document_fonts = 0` introduces some problems. Ok to merge? <sub>You can also thest against the top right icon on https://dxr.mozilla.org/mozilla-central/source/</sub>

PS. do you also have this problem? (broken ublock icons)

ublock

Thorin-Oakenpants:
@nodiscc What are your default fonts in Options>Content>Fonts&Colors>Advanced ? Do you have DejaVu Sans?

The reason I ask is that your icons are not "broken" per say. That is the padlock and power icons are showing, they are just using the wrong font. If they were broken you would see those rectangular hex symbols.

I went to a lot of trouble to day to read up on all this. It took them over 2 years to work out how to apply the patch, so excuse my ignorant less technical answer.

If your user-defined set of fonts does not return an invalid character, THEN it will allow the font (local or downloaded) - this has mostly to do with PUA.

>  give priority to the user's preferred fonts when the use_document_fonts option is off (0), but for characters not supported in the user's choice of font, use the page's specified fonts (whether local or downloaded or both) as usual, rather than going directly to system fallback fonts. This allows things like Github icons to continue working, as the PUA codepoints used should not be found in the user's pref fonts (normally), and so they'll fall back to the appropriate icon font from the page.

https://bugzilla.mozilla.org/show_bug.cgi?id=789788#c66
> The default Sans font on most Linux systems (DejaVu Sans) has glyphs for many of the PUA code points. I noticed this on github which uses this heavily all over the site. By pure dumb luck pvponline.com works because it happens to pick a dead region in DejaVu Sans.

That's the uBo issue - the icon code points in your default user defined fonts ARE valid (for the padlock and power button). The issue with hex icons being displayed is different - that's when the font is blocked, and you fall back to the invalid chars.

I hope this helps. You could try changing your default fonts in a test profile - I don't know what comes in debian, sorry. Do you have Arial - that's what windows uses for the sans.

nodiscc:
> What are your default fonts in Options>Content>Fonts&Colors>Advanced ? Do you have DejaVu Sans?

@Thorin-Oakenpants It was set to `serif` (default value)

changing it to Droid Sans, DejaVu Sans, DejaVu Serif... or the generic `sans-serif` font... fixed the uBlock incorrect icon problem as well as incorrect icons on https://dxr.mozilla.org/ http://fontawesome.io/. Thanks!! (Setting it to Arial however caused the problem to reappear).

I can never display icons properly on http://imgur.com/ regardless of my configuration. So at least this particular website problem is not linked to `browser.display.use_document_fonts`.

So my particular problem was caused by having `browser.display.use_document_fonts = 0` _and_ having a configured fallback font (`serif`) that used the same Private Use Area unicode  range as the fonts used by these pages/uBlock menu to display icons.

As the problems seems to be very specific to Debian (maybe derivatives?) there is no good reason to change the setting for everyone (it will be very hard to find a common font working on all systems). I have set `browser.display.use_document_fonts = 0` and `font.name.serif.x-western = sans-serif` in my own user.js to work around this. Closing :)

If the problem arises again we can point other users here.

Thorin-Oakenpants:
Your welcome, thanks for listening. I guess a super easy explanation is

1. Your User-Defined-Fonts [Serif?] has valid chars for a range of codepoints, hopefully not PUA ones
2. If a character is not defined in your user-defined-fonts, then it is considered invalid and treated as a glyph, hopefully these are PUA
3. It does not matter if a font is local, downloaded, or bundled with an extension
4. The pref isn't really "glyphs" but "invalid chars", it was just the simplest easiest approach to glyphs

In a perfect world, system/browser default fonts would not use PUA, and all glyphs would use PUA - but we know this is not the case, hence weird things happen

PR #313, opened 2017-06-13
[relaxed] do not disable geolocation features

nodiscc:
Geolocation attempts require user's consent through a doorhanger dialog and are fully opt-in and transparent.
Mozilla geolocation services are also used instead of Google's which lessens concerns about massive data mining/profiling capabilities.


pyllyukko:
It's still annoying and I don't see any benefit enabling it. So nope.

PR #314, opened 2017-06-13
[relaxed] don't disable password manager

nodiscc:
Add a NOTICE recommending use of the Master password feature.
`security.ask_for_password = 2` makes sure the password storage is locked again after a short interval.
Using a password manager is recommended to avoid the temptation to use simple/memorable passwords and to reuse passwords across websites. Though there are other password managers available (KeePass...), the native Firefox one is probably the easiest to use and requires no extra configuration steps (except setting a master password)

### DIFF
--- a/user.js
+++ b/user.js
@@ -742,9 +742,10 @@ user_pref("browser.cache.disk_cache_ssl",			false);
 // CIS Version 1.2.0 October 21st, 2011 2.5.5
 user_pref("browser.download.manager.retention",			0);
 
-// PREF: Disable password manager
+// PREF: Disable password manager (disabled)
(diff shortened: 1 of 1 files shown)


nodiscc:
For reference: https://github.com/Unode/firefox_decrypt - tool to extract passwords from Firefox profiles

> It can be used to recover passwords from a profile protected by a Master Password as long as the latter is known. If a profile is not protected by a Master Password, a password will still be requested but can be left blank. If the Master Password is not known it will simply fail to recover any data.

**Make sure to set a Master password to protect Firefox's password storage against basic malware that could extract your password information**

PR #337, opened 2017-09-08
add NOTICE for dom.workers.enabled

nodiscc:
Breaks some functionality on https://mega.nz/


pyllyukko:
We could also mention that it breaks <https://web.whatsapp.com/>.

pyllyukko:
I squashed this with 1ade3028e2ad58ea42bae8c5d81e23865606b8da.

Issue #338, opened 2017-09-19
Suggesting an another approach 


tunguyen6:
The problem is the user.js file causes many compatibility issues, so it cannot be deployed in mass (casual users, organizations). Therefore we would leave Firefox user.js alone or worse is that the users would use Chrome! That defeats the purpose. At personal level it also causes a lot of inconveniences. If I wanted a really high level of privacy I'd use Tor Browser instead, which leads to point number 2.
As mentioned in issue #316, with respect, I think it's a waste of time, I doubt the hardened Firefox is better than Tor Browser (with or without Tor), the more this projects progresses, the more it becomes like Tor Browser. It's not worth it.

Let's build a user.js file and other presets (such as add-ons) that balance between privacy and compatibility, and lean more to compatibility and user experience, that can be deployed in mass. This project still holds values, so what I'm suggesting is another project with that approach.
What are your thoughts?

pyllyukko:
> Let's build a user.js file and other presets (such as add-ons) that balance between privacy and compatibility, and lean more to compatibility and user experience, that can be deployed in mass. This project still holds values, so what I'm suggesting is another project with that approach.
> What are your thoughts?

The [relaxed](https://github.com/pyllyukko/user.js/tree/relaxed) branch aims to be pretty much what you described. Compatible and less breakage that is.

tunguyen6:
I didn't see that, thanks. But the download link is the same with the original version? Please update it.

tunguyen6:
I didn't go through the main page before trying it on my own Firefox profile, I lost all of my browsing history. Luckily I have a backup a few weeks ago. The relaxed version isn't that "relaxed". There is still a lot of work before I can deploy Firefox with a custom user.js.

nodiscc:
> The problem is the user.js file causes many compatibility issues

It is built as a boilerplate with every hardening option enabled, so yes. The issues should all be listed in https://github.com/pyllyukko/user.js#known-problems-and-limitations though, so it should be easy to change corresponding settings. If you find an issue not listed, please report it.

> I lost all of my browsing history

 - [ ] History is disabled as per https://github.com/pyllyukko/user.js/tree/relaxed#caching, but yes, it should also be listed in _Known problems and limitations_
[...]

PR #342, opened 2017-09-22
add notice about history/cache/storage removal

nodiscc:
Ref #338


ubone:
I strongly suggest you also put a short note in the top the user.js itself, like
`/* will nuke your history, tabs and cache (unless you comment those options) */`
I know it's mentioned in the readme but it's not easy to notice and it's easy to forget even
if you know about it ;_;

Issue #252, opened 2017-03-21
'intl.accept_languages: en-US, en' breaks web application, causes Firefox bug


nodiscc:
I noticed this after trying pyllyukko's unmodified user.js which has `user_pref("intl.accept_languages",              "en-US, en");` (whereas precedently I didn't have this value in my own user.js). This caused a bug in a web application I use (Tiny Tiny RSS): https://tt-rss.org/forum/viewtopic.php?t=4027

I checked accepted languages manually through `Preferences > Content > Languages`: no accepted languages are listed (list is blank)

When trying to add a language manually (Select language form drop-down list > Add), no language is added (the list of accepted languages stays blank) and the browser console throws an error: `preference.value is null`. I believe this is a bug in Firefox.

Setting `user_pref("intl.accept_languages",              "en-us, en");` (no capital letters) solves both problems (my application breaking, and the languages prefs being blank/unchangeable + browser console error)

 * `intl.accept_languages` value should be `en-us, en`
 * Maybe the languages dialog bug should be reported to Mozilla (@fmarier is this right?)

fmarier:
> Maybe the languages dialog bug should be reported to Mozilla (@fmarier is this right?)

It sounds like a minor bug (since you need to tweak your settings in `about:config` in order to get into a bad state), but it's probably also easy to fix (assuming it's just a matter of lowercasing the pref). So yes, it might be worth filing a bug.

nodiscc:
Thanks, reported at https://bugzilla.mozilla.org/show_bug.cgi?id=1349302

pyllyukko:
I actually tried looking into this (which way it should be) when I added it in d80e67469d0219b46836f737441006512febcf0e, but then decided on ```en-US```.

Thorin-Oakenpants:
This makes no sense to me. Changing the value to a rather unique value just because an add-on (and probably others) breaks. Default FF uses `en-US, en`. All this does is raise entropy IMO. Note: `browser.search.countryCode` and `browser.search.region` use `US`, and if you used `general.useragent.locale` it too would use capitals eg `en-US`

http://browserspy.dk/language.php 

Maybe the bug will address the issue with that add-on

nodiscc:
@Thorin-Oakenpants please re-read the issue and double check some facts:

 * There is no mention of a broken addon.
[...]

PR #355, opened 2017-12-18
Revert intl.accept_languages to "en-US, us"

nodiscc:
Reverts https://github.com/pyllyukko/user.js/pull/253

The bug reported at https://bugzilla.mozilla.org/show_bug.cgi?id=1349302 and https://github.com/pyllyukko/user.js/issues/252 was somehow reintroduced, but the other way around (a lowercase value now triggers it, the old value fixes the problem).

I'll report the new behavior on bugzilla, in the mean time let's just revert it?

Edit: I am not sure about the purpose of forcing accept-languages to en-US. There is no information about it in comments. Is it an anti-fingerprinting measure? If so one can argue that en-US would be an unusual value in most countries. It makes sense from the TBB perspective but probably not here. Move it to ignore.list with this reasoning in comments?

### DIFF
--- a/user.js
+++ b/user.js
@@ -218,7 +218,7 @@ user_pref("browser.search.geoip.url",				"");
 
 // PREF: Set Accept-Language HTTP header to en-US regardless of Firefox localization
 // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language
-user_pref("intl.accept_languages",				"en-us, en");
+user_pref("intl.accept_languages",				"en-US, en");
 
 // PREF: Don't use OS values to determine locale, force using Firefox locale setting
 // http://kb.mozillazine.org/Intl.locale.matchOS


pyllyukko:
> The bug reported at https://bugzilla.mozilla.org/show_bug.cgi?id=1349302 and #252 was somehow reintroduced, but the other way around (a lowercase value now triggers it, the old value fixes the problem).

The list works as it's expected for me with Firefox Quantum & ESR 52.5.2.

> I'll report the new behavior on bugzilla, in the mean time let's just revert it?

Needs more verification, as I wasn't able to reproduce the bug.

> Edit: I am not sure about the purpose of forcing accept-languages to en-US. There is no information about it in comments. Is it an anti-fingerprinting measure? If so one can argue that en-US would be an unusual value in most countries. It makes sense from the TBB perspective but probably not here. Move it to ignore.list with this reasoning in comments?

Fingerprinting. Lowering entropy.

nodiscc:
> The list works as it's expected for me with Firefox Quantum & ESR 52.5.2.

I have tested both values with Firefox 57 on Windows and couldn't reproduce the bug either.

The problem manifests with the value `en-us` on Firefox 52.5.2esr-1~deb9u1 in Debian stable, it could be a debian bug; there are similar [bug reports](https://bugs.debian.org/cgi-bin/pkgreport.cgi?pkg=firefox-esr;dist=unstable#_0_3_4) about [incorrect accept-language headers](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=873672).

I have test again on a fresh debian install, with a fresh profile and test both
 - original user.js
 - my modified config (different settings combinations may cause the bug)

Do both values work as expected for you, always? Or does one trigger the bug https://bugzilla.mozilla.org/show_bug.cgi?id=1349302? What OS/distributions?

nodiscc:
>  I am not sure about the purpose of forcing accept-languages to en-US
>> Fingerprinting. Lowering entropy.

Relevant links:

 * https://bugzilla.mozilla.org/show_bug.cgi?id=1039069 Warn the user that customizing the preferred language list (Accept-Language) can be used for fingerprinting ([RFC 7231](https://tools.ietf.org/html/rfc7231#section-9.7), [2616](https://www.ietf.org/rfc/rfc2616.txt))
 * https://reviewboard.mozilla.org/r/198414/diff/7 Adds a warning in language preferences dialog, and a new preference `privacy.spoof_english`:

> Request English versions of web pages for enhanced privacy

 * https://reviewboard.mozilla.org/r/178716/diff/ Patch for Firefox 59, 19 days ago, adds when `privacy.resistFingerprinting` is `true`, displays a warning about this on startup.

> Changing your language setting to English will make you more difficult to identify and enhance your privacy. Do you want to request English language versions of web pages?

[#c25](https://bugzilla.mozilla.org/show_bug.cgi?id=1039069#c25), [#c104](https://bugzilla.mozilla.org/show_bug.cgi?id=1039069#c104) raise the same question:

> Say, in Germany, most firefox users use the German version, so identifying yourself as part of a small minority [...]
> Just for the record, that's only possible while on Tor or a VPN.

There is no clear consensus yet

nodiscc:
I can't reproduce the problem anymore

TriMoon:
See https://github.com/pyllyukko/user.js/pull/253#commitcomment-27241042

PR #363, opened 2018-01-16
Added locked_user.js with lockPref as default

stablestud:
This simplifies the installation and update of the `user.js` in the system-wide configuration when used with `lockPref`.

##### OUTDATED: 
~~I included a script (`create_locked_pref.sh`) that should be run before changes to the `user.js` has been pushed upstream.~~ 
##### UPDATE:
 I implemented a pre-commit hook that can be installed with `make hooks`.
It will create `locked_user.js` with every commit.

This keeps it simple to **update** and symlink the `locked_user.js` to `/etc/firefox/syspref.js` without creating too much hassle for the user. Simply a `git pull` should do it.

What do you think about this idea? I'm open for criticism and changes to my request. :smiley_cat:


nodiscc:
I do the same thing with a [Makefile target](https://github.com/nodiscc/user.js/blob/dbu/Makefile#L6), but to replace `user_pref(` with `pref(` (for system-wide installation). I think you should leverage the existing Makefile instead of creating a new script.

You should also update the documentation about the `lockPref` install method in README.md

That is, if @pyllyukko wants to merge this.

> that should be run before changes to the user.js has been pushed upstream

You can also create pre/post-commit git [hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) to automate this.

stablestud:
@nodiscc Thanks for your comment, I re-thought the implementation.
Thanks to your tip with the `makefile` and git hooks I created now a hook which is the same as [create_locked_pref.sh](https://github.com/stablestud/user.js/blob/bf7bc620fb7f023982c40cfd765fa891ac3616ca/create_locked_pref.sh). 

To simply install the hook run `make hooks`.

Makefile
```diff
@@ -100,31 +100,29 @@ toc:
 	echo "* [$$line](#$$anchor)"; \
 	done
 
-.PHONY hooks
+HOOK = .git/hooks/pre-commit
+.PHONY: hooks
 hooks:
```
stablestud: Here starts the `make hook` implementation in the Makefile.
It creates the file `.git/hooks/pre-commit`

pyllyukko:
> I do the same thing with a Makefile target, but to replace user_pref( with pref( (for system-wide installation). I think you should/could leverage the existing Makefile instead of creating a new script.

I agree.

nodiscc:
* @pyllyukko can you add a `systemwide_user.js` target to take care of the other installation option? `'s/^user_pref/pref/'`
 * Does your Make target write to a `locked_user.js` file ? (I have never seen `$< >| $@` before; I think it's specific to Make?).
 * If yes, do you think `locked_user.js`/`systemwide_user.js` should be tracked in git? That would be one less manual step for users (some may not have Make or a proper shell installed - Windows...). And no need for git hooks either.
 * If yes, you can add it as a dependency for `make all`.
 * If yes, https://github.com/pyllyukko/user.js#system-wide-installation-all-platforms should be updated.

pyllyukko:
> @pyllyukko can you add a systemwide_user.js target to take care of the other installation option? 's/^user_pref/pref/'

Sure.

> Does your Make target write to a locked_user.js file ? (I have never seen $< >| $@ before; I think it's specific to Make?).

Yes it does. ```$<``` & ```$@``` are [automatic variables](https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html).

> If yes, do you think locked_user.js/systemwide_user.js should be tracked in git? That would be one less manual step for users (some may not have Make or a proper shell installed - Windows...). And no need for git hooks either.

No. Seems redundant to have several copies of the same file.

> If yes, you can add it as a dependency for make all.
> If yes, https://github.com/pyllyukko/user.js#system-wide-installation-all-platforms should be updated.

pyllyukko:
So I decided to just add the Makefile targets, document them, pick up the README stuff (a0791c9b4d2d08e3a8398678fb14c52c8775b0e1) from @stablestud and go with that.

I think having the other files in the repo are redundant and the Git hooks make it more complex.

Still, thanks for bringing this up, @stablestud.

PR #379, opened 2018-02-25
Added notice for first-party isolation

anthologist:
It's a fairly common addon for privacy-aware people


pyllyukko:
Thanks!

beerisgood:
Maybe it would be good to include the addon too: 
https://addons.mozilla.org/en-US/android/addon/first-party-isolation/

PR #384, opened 2018-04-05
Adds description how to use user.js on Gentoo/Archlinux

stablestud:
This pull-request extends the system-wide installation instructions for source-based Linux distributions like Gentoo or Archlinux.

It also applies to all Firefox versions that were compiled from source by the user.


user.js
```diff
@@ -1,3 +1,4 @@
+//
```
stablestud: This line must be set in `mozilla.cfg`.
I included it i `user.js` to simplify the installation for the users.

pyllyukko: Any idea why this is required? Why doesn't it work properly without it?

README.md
```diff
@@ -91,6 +91,30 @@ Copy the produced file to the Firefox installation directory. The file should be
 | Linux (Debian) | `/etc/firefox-esr/firefox-esr.js`                          |
 | OS X           | `/Applications/Firefox.app/Contents/Resources/mozilla.cfg` |
 
+#### System-wide installation for source-based distributions (e.g Archlinux, Gentoo)
+
+Genereate a file as described above.
```
pyllyukko: Typo.

user.js
stablestud: Apparently, Firefox skips the first line containing `//` on purpose, to prevent loading malicious, arbitrary JavaScript code.    

Source: [Archlinux](http://kb.mozillazine.org/Locking_preferences)
>  The parser ignores the first line of this file under the assumption that it was made unparseable on purpose "to protect the privacy of the JavaScript preferences file from loading by the browser" (code imported by [bug 89137](https://bugzilla.mozilla.org/attachment.cgi?id=61802&action=diff#mozilla/extensions/pref/autoconfig/src/nsJSConfigTriggers.cpp_sec2)). 

Source: [Mozilla's bugzilla bug 89137](https://bugzilla.mozilla.org/attachment.cgi?id=61802&action=diff#mozilla/extensions/pref/autoconfig/src/nsJSConfigTriggers.cpp_sec2)
> In order to protect the privacy of the JavaScript preferences file from loading by the browser, we make the first line unparseable by JavaScript. We must skip that line here before executing the JavaScript code.

pyllyukko:
Thanks!

Issue #430, opened 2018-10-30
twitch.tv is broken


steamp0rt:
Using `relaxed` user.js. Firefox 63.0 64-bit. Windows 10 LTSC 2019, version 17763.55. No errors in console. Audio works, but the player is just all black.

![capture](https://user-images.githubusercontent.com/25724666/47688942-8926a000-dbdf-11e8-9046-b2f785f460fa.PNG)

Addons: 
![capture](https://user-images.githubusercontent.com/25724666/47688857-21705500-dbdf-11e8-96a5-752ce35dd8cf.PNG)

Youtube works fine.

pyllyukko:
Related issue: #391

bogachenko:
why are you sure this is because of the `user.js` file but not because of the expansion or subscription for the ad blocker? did you check the `user.js` file on the new firefox profile (no extensions)?

pyllyukko:
@steamp0rt: Can you confirm it's user.js and not some addon? @bogachenko said in another issue (https://github.com/pyllyukko/user.js/pull/431#issuecomment-438423668) that it works for him and it works on my Firefox 60.3.0esr (although I'm using Linux).

steamp0rt:
Works now for some reason. Dunno why.

pyllyukko:
Good to hear that. I'll close this then.

Issue #391, opened 2018-05-17
twitch.tv videos


jlskdfjlas:
Hi, can anyone tell me what specifically I need to change so that twitch.tv works?  Thanks

nodiscc:
Are there errors in your console (`Ctrl+Shift+K`) while loading the page?

pyllyukko:
Started working for me with Firefox 60.2.1esr.

pyllyukko:
@jlskdfjlas: Is this still happening?

pyllyukko:
Taking that as a no.

PR #431, opened 2018-11-13
Updated user.js

bogachenko:
This preference breaks https://twitch.tv tested on Pale Moon and Waterfox
but on firefox 63 why it works........ eh
#430 #391

### DIFF
--- a/user.js
+++ b/user.js
@@ -17,12 +17,6 @@
 // CVE-2016-5259, CVE-2016-2812, CVE-2016-1949, CVE-2016-5287 (fixed)
 user_pref("dom.serviceWorkers.enabled",				false);
 
-// PREF: Disable Web Workers
-// https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers
-// https://www.w3schools.com/html/html5_webworkers.asp
-// NOTICE: Disabling Web Workers breaks "Download as ZIP" functionality on https://mega.nz/, WhatsApp Web, upload on https://www.virustotal.com/,  and probably others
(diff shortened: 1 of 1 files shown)


bogachenko:
I urge you @pyllyukko

Thorin-Oakenpants:
`dom.workers.enabled` was removed in FF60, so it's no longer relevant for ESR60.x or current stable (see https://bugzilla.mozilla.org/1434934)

There are actually about another 40+ prefs in the user.js that are no longer relevant

pyllyukko:
If `dom.workers.enabled` breaks Twitch, then why is it that in #430 it doesn't work with FF63 & relaxed branch (which has `dom.workers.enabled` commented out (d684068532d3845c5714dc070bbcd1f3efc86d02))?

Also, "Updated user.js" is not a sufficient commit message.

bogachenko:
Why then when disconnecting this pref, twitch breaks on 2 browsers?

pyllyukko:
> `dom.workers.enabled` was removed in FF60, so it's no longer relevant for ESR60.x or current stable (see https://bugzilla.mozilla.org/1434934)

Thanks for the link!

> There are actually about another 40+ prefs in the user.js that are no longer relevant

Yes, I know that there are obsolete settings and they still don't hurt. This has been discussed in other issues.

pyllyukko:
> upd - lol downloaded your user.js and checked on a virgin clean profile.
> on firefox 63 (on which I am now) does not break
> upd2 - but it’s worth it to disable this preference on older browsers - all

Thanks for the update. I guess we can remove it.

PR #486, opened 2020-05-14
improve Geolocation settings

gcb:
`geo.enabled=false` breaks badly-coded sites when they try to access `navigator.geolocation` directly.

A better approach to replace the functionality of `geo.enabled=false` without breaking any site, is to set `permissions.default.geo=2` which silently and automatically denies geolocation requests (just like `geo.enabled=false` would do)

`wifi.logging` doesn't do anything for the last 10 firefox verions. So removed.

### DIFF
--- a/user.js
+++ b/user.js
@@ -37,14 +37,17 @@ user_pref("dom.webaudio.enabled",				false);
 
 // PREF: Disable Location-Aware Browsing (geolocation)
 // https://www.mozilla.org/en-US/firefox/geolocation/
-user_pref("geo.enabled",					false);
+// NOTICE: Disabling geo breaks functionality on some sites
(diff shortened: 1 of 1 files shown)


pyllyukko:
> `geo.enabled=false` breaks badly-coded sites when they try to access `navigator.geolocation` directly.

Can you give an example?

gcb:
I can't link to anything as my use cases was a real state site map search, behind a login.

But a proof of concept is pretty easy, just call the API on the code. It will cause an error on the thread since the geo object will be missing, and prevent the rest of the code (in my case, global scope) from continuing execution.

After my changes, it will instead work fine, but the API will trigger a callback with the exact same side effect as if the user had pressed Deny on the UI permission dialog.

PR #487, opened 2020-05-19
Makefile: tbb-diff: update torbrowser user.js to latest version (68.80esr-9.5-1)

nodiscc:
ref. https://gitweb.torproject.org/tor-browser.git/refs/


pyllyukko:
Thanks!

PR #507, opened 2021-02-08
Split user.js into separate files

Kakadu:
And build final one using `make user.js`. At the moment it generated exactly same config as current `user.js`

This will allow us to apply fixes for the config without touching
original config. Can be useful
  * for maintainer when adding changes for the config from `master` branch
  * for users by overriding some parts of config to repair broken
  web-sites.

@pyllyukko


Kakadu:
@pyllyukko Do you have any thoughts about my idea?

pyllyukko:
> @pyllyukko Do you have any thoughts about my idea?

Sorry this has been pending. Not splitting it up at least now.

Issue #492, opened 2020-08-22
Shellcheck tests are failing


nodiscc:
https://travis-ci.org/github/pyllyukko/user.js

pyllyukko:
We need to ditch `cas.sh`, as it's not been updated in ages and is not working anymore.

nodiscc:
https://github.com/pyllyukko/user.js/pull/514

pyllyukko:
Also: " Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on."

nodiscc:
> Since June 15th, 2021, the building on travis-ci.org is ceased

Yeah. I have moved most of my projects to Gitlab CI or Github Actions for this reason. Travis as a free CI service is basically dead. Do you want me to send a pull request to switch to GH actions? For a simple Makefile-based test suite it should be quite easy (https://github.com/nodiscc/hecat/blob/master/.github/workflows/ci.yml)

pyllyukko:
> Do you want me to send a pull request to switch to GH actions?

Sure!

PR #515, opened 2021-06-18
shellcheck: ignore check SC1117: Backslash is literal

nodiscc:
- https://github.com/koalaman/shellcheck/wiki/SC1117
- SC1117: Backslash is literal in "\:". Prefer explicit escaping: "\\:".
- Ref. https://github.com/pyllyukko/user.js/issues/492


nodiscc:
Manually running `make shellcheck` still fails for me on master (shellcheck 0.5.0-3, Debian stable). @pyllyukko can you have a look?

PR #520, opened 2021-08-15
doc: add NOTICE: Disabling ServiceWorkers breaks Firefox Sync

nodiscc:


nodiscc:
Attempting to synchronize a device where service workers are disabled will cause an error in the browser console `Pairing channel error: Error: WebSocket unexpectedly closed`

pyllyukko:
Thanks!

PR #532, opened 2021-11-24
Remove VPN promotion from content blocker report

Gitoffthelawn:


nodiscc (approved):
https://en.wikipedia.org/wiki/Mozilla_VPN has some interesting external links including a security audit report. It would be nice to include

```js
// https://en.wikipedia.org/wiki/Mozilla_VPN
```

pyllyukko:
Can you add some reference links to these settings?

Gitoffthelawn:
@pyllyukko You're writing back to me, correct and not nodiscc, correct?

If so, yes, I can see what I can dig up.  This is from about a year ago (glad you are okay!), so I'll look at the bottom of the litter box for some of my notes.  ;)

pyllyukko:
> @pyllyukko You're writing back to me, correct and not nodiscc, correct?

Yes, it is me :) @nodiscc doesn't (or shouldn't! :smile: ) have access to my account.

> If so, yes, I can see what I can dig up. This is from about a year ago (glad you are okay!), so I'll look at the bottom of the litter box for some of my notes. ;)

Yeah I'm okay, thanks. Sorry for the delay. As you might have noticed, I just haven't been very active with this project.

nodiscc:
PR https://github.com/pyllyukko/user.js/pull/543 created which should disable all VPN-related ads, with references/documentation.

PR #537, opened 2022-03-28
Fixed crashing

robinlennox:
Fixed crashing issue on network.captive-portal-service.enabled


nodiscc:
I also had crashes related to this pref, but only when I manually set it to `true`. So it's unclear what the suggested change actually improves. Are you experiencing crashes with this setting set to `false`, and no crashes when set to `true` or commented out?

robinlennox:
I just commented the line and it worked. 

OS: Ubuntu
Firefox: 98.0..2

pyllyukko:
If Firefox is crashing due to some setting it should be fixed in the browser itself. Exactly as @Gitoffthelawn said. Closing this.

Gitoffthelawn:
BTW, it's really important for those experiencing this issue to report it to Mozilla at: https://bugzilla.mozilla.org/

Mozilla regularly runs thousands of tests on Firefox, and they definitely work hard to fix known crash bugs right away.  So if you are experiencing a crash, there a very good chance Mozilla doesn't know about it, and may only learn about it if you report it on Bugzilla. The only other way they can reasonably learn about it is if you have enabled full telemetry, but that doesn't always provide sufficient information to isolate a bug.  Reporting via Bugzilla is always helpful.

PR #545, opened 2023-07-05
tools: move CI to Github Actions, update documentation, clarify target names, makefile cleanup, add make help target

nodiscc:
- ref. https://github.com/pyllyukko/user.js/issues/492
- ref. https://github.com/pyllyukko/user.js/pull/544
- tested and working https://github.com/nodiscc/user.js/actions
- see commit messages


pyllyukko:
Thanks for the PR!

I need to do some minor cleanups on this, as you removed some [PHONY](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html) targets that should remain `PHONY` and made some non-`PHONY` targets into `PHONY` that shouldn't be. Also you seemed to remove some of the [automatic variables](https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html) (namely `$@`), which I would like to stay as such.

I'll check those through and get this merged soon.

pyllyukko:
Well basically every target that doesn't create a file that is the target name (e.g. `diff-tbb`) needs to be `PHONY`. Also vice versa, so if a file called `policies.json` is to be created, it shouldn't be `PHONY`. For the issue that new files aren't created it would be better to use the following trick:

```
.PHONY: FORCE
FORCE:

diff-000-tor-browser.js: FORCE
	...
```

Also I'm about to improve the `make help` target in something like this:

```
help:
	@sed -n '/^[a-z]/s/^\(.\+\):\s\+#\s\+\(.\+\)/\1\t\2/p' $(MAKEFILE_LIST) | expand -t20
```

That way we don't need to have everything as `PHONY`. If you're cool with that, I'll make the corrections and get forward with the merging?

pyllyukko:
Something like this <https://github.com/pyllyukko/user.js/tree/minor-fixes-to-545>?

nodiscc:
> Well basically every target that doesn't create a file that is the target name (e.g. diff-tbb) needs to be PHONY

:+1: 

> For the issue that new files aren't created it would be better to use the following trick:

I didn't know this trick: https://www.gnu.org/software/make/manual/html_node/Force-Targets.html. Thanks! Not sure about the benefits compared to just using `.PHONY` though (maybe cleaner/more correct, as `.PHONY` implies the target does not create a file with the same name?), but it works for me.

> improve the make help target in something like this:

Perfect. Tested https://github.com/pyllyukko/user.js/blob/minor-fixes-to-545/Makefile, works fine.

nodiscc:
Closed accidentally, feel free to reopen.

pyllyukko:
What...?

nodiscc:
I'm sorry, my gitea instance is misbehaving. The branch `cleanup-tests` has been deleted on it, and it is configured to mirror (push) to https://github.com/nodiscc/user/.js, so the branch gets deleted again on github/gitlab every time the mirror synchronization runs (every few hours)...

image

I will disable mirroring for now. Sorry for the noise.

Issue #550, opened 2024-02-10
set `network.http.referer.trimmingPolicy = 2`/`network.http.referer.XOriginTrimmingPolicy = 2`


nodiscc:
While reviewing https://wiki.mozilla.org/Privacy/Privacy_Task_Force/firefox_about_config_privacy_tweeks, I found these prefs which are currently not set in `user.js`

` network.http.referer.trimmingPolicy = 2`

> Send only the scheme, host, and port in the Referer header

       0 = Send the full URL in the Referer header
       1 = Send the URL without its query string in the Referer header
       2 = Send only the scheme, host, and port in the Referer header

` network.http.referer.XOriginTrimmingPolicy = 2`

> When sending Referer across origins, only send scheme, host, and port in the Referer header of cross-origin requests. Source

       0 = Send full url in Referer
       1 = Send url without query string in Referer
       2 = Only send scheme, host, and port in Referer
 

There is a comment in user.js about this pref:

```
TODO: https://github.com/pyllyukko/user.js/issues/94, commented-out XOriginPolicy/XOriginTrimmingPolicy = 2 prefs
```

This and #549 should cover all settings detailed on the mozilla wiki page.

pyllyukko:
Sounds like something we want to add. Want to make a PR for this?

nodiscc:
Sure, putting this on my list of things to do, expect a PR sometime soon.

Issue #94, opened 2016-01-02
network.http.referer.spoofSource and browser.privatebrowsing.autostart breaks sites.


alabrand:
network.http.referer.spoofSource, for example, breaks Bato.to and some other sites.
browser.privatebrowsing.autostart, for example, breaks 4chan and a few other sites.

I suggest leaving these to their default value but encouraging the use of them on the main github page if at all possible.

Hunter-Github:
How many people willingly use the sites you cited while knowing about the privacy implications?

pyllyukko:
> network.http.referer.spoofSource, for example, breaks Bato.to and some other sites.
> browser.privatebrowsing.autostart, for example, breaks 4chan and a few other sites.

Make sure your ```network.http.sendRefererHeader``` setting value is ```2```. I haven't seen any sites that break with this setting combination. HTTP referer header is one of the worst information leakages IMO, so we should really keep the spoofing turned on.

How exactly does ```browser.privatebrowsing.autostart``` break 4chan and these other sites?

> I suggest leaving these to their default value but encouraging the use of them on the main github page if at all possible.

I'm afraid it's not possible as these settings are really important regarding privacy.

alabrand:
Hm, am I doing something wrong then? It's a clean Firefox installation, and the default value for network.http.sendRefererHeader is 2 so that shouldn't be a problem.

Private Browsing breaks 4chan in that you can't use certain functions. Or actually, when I think about it, that could have been referer spoofer breaking it instead of Private Browsing.


Still, I don't know why referer spoofer didn't work. As I said, the default value on a clean Firefox installation for network.http.sendRefererHeader is 2.

pyllyukko:
> Private Browsing breaks 4chan in that you can't use certain functions. Or actually, when I think about it, that could have been referer spoofer breaking it instead of Private Browsing.

It is more likely, yes. I can't think of any reason, why private browsing would break anything like that. Can you say what exact functions does it break?

> Still, I don't know why referer spoofer didn't work. As I said, the default value on a clean Firefox installation for network.http.sendRefererHeader is 2.

I guess the web application just has strict checking against the referer header and spoofing doesn't cut it.

pyllyukko:
Relates to #2

Atavic:
[...]

PR #554, opened 2024-02-12
Trim HTTP referers

nodiscc:
- fixes https://github.com/pyllyukko/user.js/issues/550

Same warning as for https://github.com/pyllyukko/user.js/pull/553, altering referers is a frequent cause for breakage on sites that rely on them (I didn't include the notice because it is already present twice in prefs above).

For a relaxed variant I recommend first relaxing `network.http.referer.trimmingPolicy` to 1 then 0 (because `network.http.referer.XOriginTrimmingPolicy` still protects against sending full referers across origins), and in last resort relaxing `network.http.referer.XOriginTrimmingPolicy`.

Again, these prefs *will* break websites, but the notice is present.


nodiscc:
See also https://github.com/pyllyukko/user.js/issues/94#issuecomment-256515175

>I was thinking that XOriginPolicy = 2 and XOriginTrimmingPolicy = 2 might be an appropriate default that will not break very much. People can comment out XOriginPolicy if they see breakage they can't live without.

>I'm not entirely convinced that TrimmingPolicy does anything useful over XOriginTrimmingPolicy because you can't hide your query string from the server you're connecting to. They have all of your traffic in their access log and it's easy to link all of your page views together. They don't need the referrer to do any of that.

PR #558, opened 2025-01-04
Readme: Remove HTTPS Everywhere add-on

Croydon:
The add-on isn't maintained anymore for years.

It also isn't necessary anymore, since Firefox now has the `dom.security.https_only_mode` config to enable a HTTPS-only mode, and since https is nowadays pretty much the default anyway.


pyllyukko:
Thanks!

